PR #1, opened 2015-05-12
Jamie capocaccia2015

neworderofjamie:
I was trying to add some new models at Capo Caccia and I was getting really annoyed with the application magic numbers - What do you think of this MD5-based replacement? This change irritatingly crosses into sPyNNaker and sPyNNakerExtraModels as well so will create more pull requests


rowleya:
Although the magic number bit is due to change, the change here is still useful as is, and avoids the user having to think about this, and the other changes are also useful in the associated projects.

PR #13, opened 2015-07-21
Speed up data transfer 2

rowleya:


spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -16,6 +16,6 @@ def read_and_convert(x, y, address, length, data_format, transceiver):
     """
 
     # turn byte array into str for unpack to work
-    data = str(list(transceiver.read_memory(x, y, address, length))[0])
-    result = struct.unpack(data_format, data)[0]
+    data = transceiver.read_memory(x, y, address, length)
+    result = struct.unpack_from(data_format, data)
```
alan-stokes: we should keep the [0] from the result line. the only calls to this end up doing it there, which makes it a bit ugly. Having it here simplies the code all around.

rowleya: The reason that I took this out was that the format is just a string, and thus can be anything.  If it is a single character, using [0] is fine, but if you are decoding a number of fields, you won’t be able to if you push the [0] in here, which could be confusing.  Maybe that is part of the point of this function, but maybe it can be changed to be read_int, and then the format goes away (I think that the function is only ever used in this context).

PR #14, opened 2015-07-29
added some provanence data gathering from the live packet gatherer an…

alan-stokes:
…d cleaned up the code slightly via the use of enums and adding a provanence data region to the live packet gather.


c_common/models/live_packet_gather/live_packet_gather.c
```diff
@@ -22,6 +22,12 @@ static circular_buffer without_payload_buffer;
 static circular_buffer with_payload_buffer;
 static bool processing_events = false;
 
+//! Provanence data store
+static uint32_t number_of_over_flows_none_payload = 0;
+static uint32_t number_of_over_flows_payload = 0;
+static uint32_t NUMBER_OF_PROVANENCE_DATA_ELEMENTS = 2;
+static uint32_t SIZE_OF_PROVANENCE_ELEMENTS_IN_BYTES = 8;
```
neworderofjamie: How come the naming of the last two is inconsistent? Also, it might make sense to put these in a struct - Then you wouldn't need the enum as well AND the code to copy it back into SDRAM would be simpler

c_common/models/live_packet_gather/live_packet_gather.c
```diff
@@ -119,6 +156,27 @@ void flush_events(void) {
...
+    provanence_data[NUMBER_OF_LOSS_PACKETS_WITH_PAYLOAD] =
+        number_of_over_flows_payload;
+    // Copy provanence data into sdram region
+    memcpy(provanence_region_address, provanence_data,
+           SIZE_OF_PROVANENCE_ELEMENTS_IN_BYTES);
+    log_info("The provanence data consisting of %d lost packets without "
+             "payload and %d lost packets with payload.",
+             number_of_over_flows_none_payload, number_of_over_flows_payload);
+}
+
```
neworderofjamie: As I said above, this seems overly circuitous! Either using a struct here or just copying straight into provanence_region_address would be be better.

spinn_front_end_common/abstract_models/abstract_provides_provanence_data.py
```diff
@@ -12,11 +12,15 @@ def __init__(self):
         pass
 
     @abstractmethod
-    def _write_provanence_data_in_xml(self, file_path):
+    def write_provanence_data_in_xml(self, file_path, transciever,
+                                     placement=None):
         """
         abstract method to force objects extending this to provide a xml
         element object for output
         :param file_path: the file apth to write the provanence data to
```
neworderofjamie: 'path' and 'provenance' :)

c_common/models/live_packet_gather/live_packet_gather.c
patrickcamilleri: Do you guys mean Provenance?

c_common/models/live_packet_gather/live_packet_gather.c
```diff
@@ -22,6 +22,12 @@ static circular_buffer without_payload_buffer;
 static circular_buffer with_payload_buffer;
 static bool processing_events = false;
 
+//! Provanence data store
+static uint32_t number_of_over_flows_none_payload = 0;
+static uint32_t number_of_over_flows_payload = 0;
```
neworderofjamie: It might make sense to put the provenance data fields either:

a) In a struct 
b) In an array of uint32_ts indexed with the enum

A struct would have the benefit of removing the need for an enum and both options would simplify the code to copy it back into SDRAM.

alan-stokes:
Hi Jamie, Patrick,

Im happy to move it into a struct and make the movement to sdram cleaner. Was just a quick build and I personally find enums better to maintain than fixed ints spread all over the code (as it was previosuly). its good you two caught and had a better plan up your sleaves. Once ive done Jamies possien fix, ill get onto it.

@jamie what do you mean by "'path', 'provenance' and 'transceiver' :)"? as these parameters are the path to the xml file to write it in, the spinnman interface object and the placement object for the subvertex, so that it knows where to read its data from. If you wanted to access to the provanence data directly, we'll need to redeisgn this interface to return provanence, instead of just writing it to a file.

neworderofjamie:
They're all spelt wrong :)

alan-stokes:
ah. ok. dyslesica fail for myself.
Will try to fix. lol.

alan-stokes:
Hi Jamie, wnat to verify the changes and approve the pull request please.
thank you

alan-stokes:
oops. Missed that one, had all the python changed. lol

this has not been necessary for more years than, even I, have been alive!????????

neworderofjamie:
All those unnecessary voids - Probably haven't been required since @dr-david-lester first bothered heavy machinery using C :)

PR #15, opened 2015-07-29
added support for infinite running and tested with a modified version…

alan-stokes:
sorts out infinite runs and thus alows the push bot pull request to be compelted

### DIFF
--- a/spinn_front_end_common/abstract_models/abstract_data_specable_vertex.py
+++ b/spinn_front_end_common/abstract_models/abstract_data_specable_vertex.py
@@ -1,7 +1,11 @@
-from data_specification.enums.data_type import DataType
+"""
+AbstractDataSpecableVertex
+"""
 from data_specification.file_data_writer import FileDataWriter
 
 from spinn_front_end_common.utilities import exceptions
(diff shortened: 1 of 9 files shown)


spinn_front_end_common/abstract_models/abstract_data_specable_vertex.py
```diff
@@ -39,7 +43,14 @@ def _write_basic_setup_info(self, spec, region_id):
...
         spec.write_value(data=int(application_name_hash, 16))
         spec.write_value(data=self._machine_time_step * self._timescale_factor)
-        spec.write_value(data=self._no_machine_time_steps)
+        if self._no_machine_time_steps is None:
+            spec.write_value(
+                data=front_end_common_constants.INFINITE_RUN.TRUE.value)
+            spec.write_value(data=0)
+        else:
+            spec.write_value(
+                data=front_end_common_constants.INFINITE_RUN.FALSE.value)
```
neworderofjamie: This seems a little over-engineered - This is the only place in Python where this is used so why not write 0 or 1!

neworderofjamie:
Looks good

PR #29, opened 2015-11-10
Work flow multiple runs

alan-stokes:
brings in the pause and resume modes of the tool chain from felix work into the tool chain proper. Works in the brnach, apart from changing the application graph mid runs.


alan-stokes:
closing as due to fixes in lower brnach contains these

PR #52, opened 2016-06-17
Reverse ip tag partitioned fixes

rowleya:
Allows reverse ip tag multicast source to work as a partitioned vertex more smoothly.
 - [x] Wait for merge of #51

### DIFF
--- a/c_common/front_end_common_lib/Makefile.SpiNNFrontEndCommon
+++ b/c_common/front_end_common_lib/Makefile.SpiNNFrontEndCommon
@@ -28,14 +28,12 @@ $(foreach dir, $(SOURCE_DIRS), $(eval $(call define-build-code,$(dir))))
 OBJECTS += $(OBJS)
 
 LIBRARIES += -lspinn_frontend_common -lspinn_common -lm
-ifndef DEBUG
-    DEBUG = PRODUCTION_CODE
-endif
+FEC_DEBUG := PRODUCTION_CODE
 # Run md5sum on application name and extract first 8 bytes
 SHELL = bash
 APPLICATION_NAME_HASH = $(shell echo -n "$(APP)" | md5sum | cut -c 1-8)
 
-CFLAGS += -Wall -Wextra -D$(DEBUG) $(OTIME) -DAPPLICATION_NAME_HASH=0x$(APPLICATION_NAME_HASH)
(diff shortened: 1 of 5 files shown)


alan-stokes:
even without my change. this doesn’t work for when the RIPTMCS is used as a injector (SpikeIO). Will look at fixing, but the branch as is/was just doesn’t work. This is due to errors such as 
"
AttributeError: 'ReverseIPTagMulticastSourcePartitionedVertex' object has no attribute '_send_buffer_max_space'
"
This is because there are params that just don’t exist in a injection mode over a SSA mode. (shockingly).  Add it to the list of things wrong with RIPTMCS which hopefully wil be fixed with Alex's verison.

AlexRast:
On 08/07/16 21:44, Alan Stokes wrote:
> even without my change. this doesn’t work for when the RIPTMCS is used as a
> injector (SpikeIO). Will look at fixing, but the branch as is/was just doesn’t
> work. This is due to errors such as
> "
> AttributeError: 'ReverseIPTagMulticastSourcePartitionedVertex' object has no
> attribute '_send_buffer_max_space'
> "
> This is because there are params that just don’t exist in a injection mode over
> a SSA mode. (shockingly). Add it to the list of things wrong with RIPTMCS which
> hopefully wil be fixed with Alex's verison.

I expect all of these will be fixed; certainly there will be one universal 
parameter list for all cases, but this points out something that I've noticed 
for some time between SpikeSourceArrays and SpikeInjectors. The hope had been 
(in previous versions) that the needs of both type of use could be solved by a 
single interface - under the justification that they essentially needed similar 
things: external injection of spikes from a remote source.

The more I've looked into this and experimented with it, though, the more I am 
convinced that these are quite radically different use cases, at least in the 
typical situation. An SSA is pre-injecting large numbers of spikes for future 
injection onto the fabric: you have a traffic pattern that sees large blocks of 
(timestamped) spikes arriving in dense bursts. An SI meanwhile injects spikes 
probably most of which will be injected immediately onto the fabric with no 
further delay; spikes arrive relatively sparsely but the injection rate is 
fairly uniform. These are very different traffic statistics.

The current version I'm working on will handle both cases, but I wonder whether 
the decision to merge both use cases together shouldn't be revisited?

Meanwhile it would help to gather together a list of all the outstanding 
external I/O issues that we should fix with the upcoming (PRAERIE) version so 
that I can make sure the implementation addresses them.

PR #67, opened 2016-09-21
Fixed a timing issue when run is called multiple times

rowleya:


alan-stokes:
no idea what milestone you want for these. might be worth having a bug_fix feature. lol.

PR #77, opened 2016-10-14
Add memory file interface

rowleya:
Adds support for the memory io of SpiNNMan via an interface and algorithm which is now added by default to the workflow.  Note that the algorithm will do nothing if there are no vertices which implement the interface.
- [ ] Wait for merge of SpiNNakerManchester/SpiNNMan#58


rowleya:
This is ready to go as well (with the restriction that it depends on the SpiNNMan equivalent branch)

PR #82, opened 2016-11-02
Report fix

alan-stokes:
supports running of the same script multiple times without  having to wait for the first to finish. This has resulted in the loss of the latest folder, but the timestamps of the reports are still accurate and so the top/bottom one of a file browser should show the correct folder.


alan-stokes:
closed due to further one with more changes.

PR #83, opened 2016-11-07
Working pushbot

alan-stokes:
fixes: 
fixing the command sender and external device interface so that it works with the shed load of new features. Its resulted in many different features being added on top of it and eventually breaking it, but without being tested, they haven't been picked up till now. 

modules this fix is associated with 
1. SPY
2. PACMAN
3. SpyExternalDevices
4. SPINNMan
5. PyNN examples
6. GFE 

review these before this one:
- [x] PACMAN https://github.com/SpiNNakerManchester/PACMAN/pull/53
- [x] SpiNNMan https://github.com/SpiNNakerManchester/SpiNNMan/pull/61
- [x] Revert SpiNNMan to master in .travis.yml


rowleya:
Prior to merging, please make sure that SpiNNMan links to master in travis.yml

alan-stokes:
note for rowley. extra models dont compile with these branches. pull request merge is on hold whilst black magic is recasted. but will review rest anyhows

rowleya:
In a discussion today, we thought we might be able to add the commands to the command sender after key allocation to allow the assigned retina key to be put in the payload of a command.

Unfortunately, I don't think this will work.  This is because the command sender needs to know the keys that the commands are going to use in order to build partition identifiers prior to key allocation (the command keys are known in advance at this point).  Thus I would suggest this is raised as an issue to possibly be resolved in the future by some other mechanism.  Otherwise, this isn't going to be merged in time for a release by Capocaccia.

PR #84, opened 2016-11-21
app data latest issue

alan-stokes:
when debugging, if mapping fails for any reason, the next run would cause folder generation to fail. switching the two tests around fixed it. This is likely made redundant by the multi-run fixes, but is still needed for current version.


rowleya:
Seems to have been made redundant as predicted, so deleting

PR #100, opened 2017-02-24
Fix iobuf extraction

rowleya:
Though the title suggests this branch fixes the iobuf extraction, it also fixes a number of other issues:
 - Fixes the extraction of buffers on second and subsequent runs
 - Adds IOBUF clearing on second and subsequent runs rather than clearing post-run
 - Simplifies the IOBUF and Provenance extraction to be done directly after run, or on stop if the run is infinite
 - Fixes to the error handling system that make it more smooth
 - Fixes to some of the algorithms so that they check each binary uses simulation.h/c before trying to talk to it


rowleya:
This is now part of "router_compressor_on_chip" as it was doing similar things already (in a different way)

PR #104, opened 2017-03-16
Move the injection point entirely into the support class.

dkfellows:
The aim is that code that uses this class should not have to know anything about the injection mechanism at all. By factoring out, the smarts can be more cleanly restricted to the core of the Spinnaker python code, rather than spraying it into places that are more user-visible.


dkfellows:
Just because we have a complicated injection mechanism, it doesn't mean that we need to make users look at the details of it unless they insist on poking their noses in.

rowleya:
Interested to hear what you think of this solution - this makes the injection stateless.  I also renamed the property to be compatible with existing users.

dkfellows:
No longer meaningful.

PR #108, opened 2017-03-21
Allow installation from branches other than master as a coordinated set

dkfellows:
If you've got several branches that work together on a problem, this makes it easy to build them together. The magic is based on the `TRAVIS_BRANCH` environment variable and some shell fallbacks.


dkfellows:
I'm proposing to roll this sort of thing out to all repositories, of course.

PR #112, opened 2017-03-27
Fix error recovery

rowleya:
Fixes a couple of issues:
 - The error recovery was calling functions with the wrong arguments
 - Made the application runner wait for the correct amount of time when there is a timescale_factor
 - Allow router provenance to fail on some chips but continue


alan-stokes:
what does the provenance do to reflect the failure of getting router prov on specific chips?

rowleya:
Nothing at present - this could be made a new issue though.

PR #123, opened 2017-04-12
Fix critical bug in LPG caused by mixup between & and &&

dkfellows:
Tests of a bit flag must always be done with `&` and not `&&`; the latter has very different semantics indeed! To help prevent this sort of problem in future, I've defined a few helpful simple macros that describe how exactly to test a bit flag, and also how to clamp integers to their low 8 and 16 bits (because these are common and areas that it is easy to make the same sort of mistake in).

I'd feel happier giving the bits names instead of using `0x1` and `0x2`, but I don't know 'em and that's not critical to this fix.


c_common/models/live_packet_gather/live_packet_gather.c
```diff
@@ -111,19 +122,19 @@ void flush_events(void) {
...
-            sdp_msg_aer_header[0] |= (event_count & 0xff);
+            sdp_msg_aer_header[0] |= CLAMP8(event_count);
 
             g_event_message.length = sizeof(sdp_hdr_t) + header_len
                                      + event_count * event_size;
 
             if (payload_apply_prefix && payload_timestamp) {
                 uint16_t *temp = (uint16_t *) sdp_msg_aer_payload_prefix;
 
-                if (!(packet_type && 0x2)) {
```
dkfellows: This is the place where the bug actually is. fixed.

alan-stokes: On the naming of the EIEIO header bits, heres the doc that covers it. 

http://spinnakermanchester.github.io/2015.004.LittleRascal/EIEIOProtocolProposal.pdf

0x2 is F (or packet format (command or basic)
0x1 is P (with or without key prefix, which is its attempt of compression)

Hope that helps. That whole code and its corrasponding RIPMCS needs a overall in all honestly.

PR #124, opened 2017-04-13
Enforce all boards to use the same buffer manager listener port

oliverrhodes:
Together with the changes in SpiNNMan under the branch of the same name (additional_listeners_removal), this should fix the issues around the buffer manager creating multiple listeners on the same connection.


oliverrhodes:
Added an integration test to check listener creation, so this branch should now be ready to merge.

Note that the corresponding branch in SpiNNMan should be merged first (this should cause all Travis checks to pass, and not just the 'push' build)

PR #140, opened 2017-06-05
attempts to fix neo segments

alan-stokes:
fixes neo segments for recording over multiple runs. 

this one just allows buffer extractor to run before front end calls as needed. 

depends upon:
Spynnaker8 https://github.com/SpiNNakerManchester/sPyNNaker8/pull/8
spynnaker https://github.com/SpiNNakerManchester/sPyNNaker/pull/320


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1415,21 +1415,23 @@ def _do_run(self, n_machine_time_steps, loading_done):
...
         if (self._has_ran and not self._has_reset_last and
                 not self._use_virtual_board):
             algorithms.append("FrontEndCommonBufferExtractor")
 
-            # check if we need to clear the iobuf during runs
-            if self._config.getboolean("Reports", "clear_iobuf_during_run"):
-                algorithms.append("FrontEndCommonChipIOBufClearer")
+        # update algorithm list with extra pre algorithms if needed
+        if self._extra_pre_run_algorithms is not None:
+            algorithms = list(self._extra_pre_run_algorithms)
```
dkfellows: I can't trust that that code will work; we don't want to wipe out the algorithm we potentially added above. I believe that line 1428 should instead be:
```python
algorithms.extend(self._extra_pre_run_algorithms)
```

alan-stokes:
closed as fixed in networks pull request

PR #141, opened 2017-06-13
fixes for config and getboolean

alan-stokes:
fixes issues with config.get and config.getboolean

p.s. we should build some simple tests to ensure this wont hit us again.

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1315,8 +1315,8 @@ class AbstractSpinnakerBase(SimulatorInterface):
         outputs = []
         algorithms = [self._dsg_algorithm]
 
-        if (self._config.get("Reports", "reportsEnabled") and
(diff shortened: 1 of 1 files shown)


rowleya:
Should really have a test, but I can see how hard such a test would be and that it wouldn't actually be of much use...

alan-stokes:
thats why i plonked it as a comment. i imagine christian will take the challenge and figure a way of detecting it all. It is what he does afterall. lol

rowleya:
Though I am happy with a test, I can't see a test helping us to not make this mistake again in the future elsewhere in the code.  The only test would be that provenance data gathering is done when the option is turned on I guess...

rowleya:
Or else that it is not done when turned off

alan-stokes:
talking of that line of stuff, theres a issue with reports for multiple calls to run. it seems (from peters version) that p.run(x) p.run(y) does not produce two run folders. it plonks them into the same folder in run1, and uses the unqiue identifier to get around the issue that the file already exists. im opening up a issue for it so that we can chase this down in the future

PR #145, opened 2017-06-14
Profiler update (renamed)

rowleya:
Some changes moved from SpiNNakerManchester/sPyNNaker#196
- [x] Move profiler code here
- [x] Update profiler interface to include a data structure, interface and some utilities
- [x] Add an algorithm to extract profile data and display the results when available
- [x] Wait for merge of #50 
- [ ] Wait for merge of #163


alan-stokes:
the gather has not yet been tied into the workflow, and has obviously not been tested in isolation, given that the interface call requires transceiver and placements, and yet the gather only ever uses transceiver and no injection. 

will fix and try to tie everything in

alan-stokes:
this is now working. There is a belief that we could tie this into the recording region and auto pause and resume. But the complexity of figuring out how often this is called per profile method is rather complex and therefore left as future work if it ever becomes criticle to do so.

rowleya:
Note that this now needs to wait for another FEC merge, as the data generation is basically broken in master

PR #146, opened 2017-06-20
Recording dma fix

rowleya:
Fixes DMA-based recording.  In particular:
 - DMA recording is now explicitly requested by the API user; if they don't request it, they get "normal" (i.e. using spin1_memcpy) recording.
 - DMA recording accepts a callback (if this is provided as NULL, you get normal recording).
 - The callback priority for recording with DMA needs to be 0 or less if a "wait" is going to be done for the recording (this is up to the API user of course).

### DIFF
--- a/c_common/front_end_common_lib/include/recording.h
+++ b/c_common/front_end_common_lib/include/recording.h
@@ -14,6 +14,9 @@
 
 #define RECORDING_DMA_COMPLETE_TAG_ID 15
 
+//! \brief Callback for recording completion.
+typedef void (*recording_complete_callback_t) ();
+
 typedef struct {
     uint16_t eieio_header_command;
     uint16_t chip_id;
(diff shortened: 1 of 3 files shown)


alan-stokes:
@andrewgait what you doing merging this in whilst there were unclosed discussions?! please revert master and reopen this pull request. I personally think this fix is clunky and can be fixed with a much smaller changes.

alan-stokes:
ok, so i missed the summit button. so your not completely wrong, but rowley asked me to review it, and i was doing so before lunch.  i have misgivings about this work, and now i have no chance to change it before going into master.

rowleya:
If there are interface cleanings to be done, feel free to create a new branch...

alan-stokes:
lol. add it to the pile, ive got way too many things on my plate right now. how about lobbing it over to someone who has less workload to fix.

andrewgait:
I can have a look at it next week if no-one else has got to it before then.

PR #148, opened 2017-06-21
fix by Alan to getting binaries when using virtual machine

Christian-B:


Christian-B:
If this pull request is Accepted the branch
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/tree/more_integration_tests

can also be deleted as they are the same duplicated for travis

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -719,6 +714,12 @@ def run(self, run_time):
                 self._do_load()
                 loading_done = True
 
+        # Disable auto pause and resume if the binary can't do it
+        if (self._executable_start_type !=
+                ExecutableStartType.USES_SIMULATION_INTERFACE):
+            self._config.set("Buffers", "use_auto_pause_and_resume",
+                             "False")
```
rowleya: Not understanding why this has moved... surely this breaks things since the check for auto pause and resume has already been done...

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1369,16 +1363,23 @@ def _do_load(self):
         # algorithms needed for loading the binaries to the SpiNNaker machine
         optional_algorithms.append("FrontEndCommonLoadExecutableImages")
 
+        # Get the executable targets
+        optional_algorithms.append("FrontEndCommonGraphBinaryGatherer")
```
rowleya: Should this come before the loading of the executables?  I know that the executor can reorder them (and probably will), but it saves my brain from hurting so much if it is in approximately the right order to start with...

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: alan says: if its where it originally was, then itll never be set, as the executable start type is now only set at load time. 

theres a equiv of thiswhich had to have a is not virtual board due to the same interaction with multirun

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: if done before loading, then your virutal board needs to find them. by being done at loading the virtual board can work, and as the executables are only found jsut before they are loaded, they might as well be in loading. 

i dont mind moving them about, we just put them at the first poitn where optional algorithms was built.

Christian-B:
I was able to run various intergartion tests after this fix including:
https://github.com/SpiNNakerManchester/sPyNNaker8/tree/more_integration_tests/p8_integration_tests/test_auto_pause_and_resume_tests

So either it works with auto pause resume or that test is not good enough.

rowleya:
It's a good test for sPyNNaker, but the GFE will be where the issue could occur, since that is more likely to have a binary that isn't a Simulation binary.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: OK, so then this is not the correct fix ;).  The issues are:

1. The check on line 666 will be invalid.  This will allow it to use auto pause and resume when the binary isn't a simulation binary, which won't work.
1. Executable start type won't be set at all when using a virtual board.  This seems dangerous (although it looks like it will get the right answer)

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: I just mean, swap line 1364 with this one (and the comments), so that the order is correct (gather then load).

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: 2. executable start type has no position in a virtual board run, thus not being set is the right thing to do. your not going through the loading or running phase, so no point setting a param for no reason. separation of concerns and all that. To set it, you need to find the binaries, which is what were trying to avoid when using a virtual board. 

1. ill need to look at in my ide, as trying to traverse the gituhb equiv is not helping. will look at after the dma review

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: dont mind here. as stated before. the position in the list was not what we were disucssing, more the move into do_load

alan-stokes:
fixed issues. please review. 


p.s. me hungry......very very hungry

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1366,19 +1364,23 @@ def _do_load(self):
...
             "LoadedReverseIPTagsToken", "LoadedIPTagsToken",
             "LoadedRoutingTablesToken", "LoadBinariesToken",
-            "LoadedApplicationDataToken"
+            "LoadedApplicationDataToken", "ExecutableTargets"
         ]
 
         executor = self._run_algorithms(
             inputs, algorithms, outputs, optional_algorithms)
         self._load_outputs = executor.get_items()
+        self._executable_start_type = executor.get_item("ExecutableStartType")
```
rowleya: Guessing you don't really need to do this anymore...

spinn_front_end_common/interface/interface_functions/front_end_common_locate_executable_start_types.py
```diff
@@ -0,0 +1,38 @@
+from spinn_front_end_common.utilities import exceptions
+from spinn_utilities.progress_bar import ProgressBar
+
+
+class FrontEndCommonLocateExecutableStartType(object):
+
+    def __call__(
+            self, placements, graph, executable_finder, graph_mapper=None):
```
rowleya: I suspect this algorithm doesn't need all these inputs - it could probably operate simply on the application / machine graph.

spinn_front_end_common/interface/interface_functions/front_end_common_locate_executable_start_types.py
```diff
@@ -0,0 +1,38 @@
...
+
+        progress = ProgressBar(
+            graph.n_vertices, "Finding executable_start_types")
+        binary_start_type = None
+        for vertex in progress.over(graph.vertices):
+            placement = placements.get_placement_of_vertex(vertex)
+            associated_vertex = graph_mapper.get_application_vertex(vertex)
+            placement_binary_start_type = \
+                associated_vertex.get_binary_start_type()
```
rowleya: Not all vertices have this function - should be checking for isinstance AbstractHasAssociatedBinary

rowleya:
Note that a unit test is now also broken...

spinn_front_end_common/interface/interface_functions/front_end_common_locate_executable_start_types.py
alan-stokes: the orignal code didnt check either then. lol. i copied it directly from the binary gatherer. lol

rowleya: Oh it did - take another look :)

alan-stokes: *hat off*. put in now

spinn_front_end_common/interface/interface_functions/front_end_common_locate_executable_start_types.py
alan-stokes: done. still needs graph mapper and ive taken the message to be vertex label instead of the placement to remvoe that requirement

rowleya:
Still needs flake8 fixes

alan-stokes:
flaked and baked

PR #163, opened 2017-07-06
Allow uncompressed

alan-stokes:
Allows regions to be left uncompressed. Allows the reservation of space in a region which is then not used directly, but is needed later.

- [ ] Wait for merge of SpiNNakerManchester/DataSpecification#24


rowleya:
This should now work

rowleya:
Tests added to be sure

PR #164, opened 2017-07-17
checks Is running

Christian-B:


Christian-B:
Current version of PyNN8 does not allow run(None) as run passes through PyNN which can not accept None.
So this is not required there.

GraphFrontEnd is also not done yet but can be added at any time.

spinn_front_end_common/utilities/globals_variables.py
```diff
@@ -13,6 +13,17 @@ def get_simulator():
     return _simulator
 
 
+def get_not_running_simulator():
+    global _simulator, _failed_state
```
rowleya: It looks like this just does the same as get_simulator, with an additional check to verify that the simulator isn't running.  Can this instead call get_simulator rather than repeating the code?

PR #171, opened 2017-08-01
reports incompatible with run forever

Christian-B:


alan-stokes:
bounced back so that its more future proof. discussed with christian directly

PR #174, opened 2017-08-02
skip energy report if no buffer manager (stop called after crash)

Christian-B:


alan-stokes:
wrong place to do this check. fix at abstract spinnaker base

PR #176, opened 2017-08-02
emergency fix for getting GFE working.

alan-stokes:
fixes for getting gfe working with chip-power-monitor

needs corrasponding gfe thing.


alan-stokes:
monitors is done by spinn manchines call. injectors are monitors.

PR #181, opened 2017-08-03
Rewrite of database access layer

dkfellows:
This is a rewrite of the database access layer to make the accesses not vulnerable to SQL injection attacks. It also moves the definitions of the tables into a separate file (`…/db.sql`) so that they are much easier to understand as a whole.

It also takes care to avoid quadratic behaviour when inserting values into the database by tracking more carefully the mapping from vertices and edges to their database IDs.


andrewgait:
Looks fine to me.

PR #184, opened 2017-08-10
Avoid mode debug setting zero to True

Christian-B:


alan-stokes:
this would be a problem for all get_booleans. I dont the reasoning for treating this one differently.

Christian-B:
Went for more of a harc code in Options approach as discussed.

rowleya:
Seems to have a Flake8 Error.  This is marked as for merging before release.  Is this still desired?

Christian-B:
Yes this needs merging ASAP or Mode debug will always fail!

Issue #63, opened 2016-09-07
Suppress printing of INFO: Starting new HTTP connection (1)


rowleya:
This is caused when using the HBP portal when checking that the allocated machine is still available

rowleya:
Fixed in #187

PR #187, opened 2017-08-15
Lower requests logging level

rowleya:
Lowers the logging level of requests library to suppress a message displayed on the portal


rowleya:
Fixes #63

dkfellows:
Looks OK to me; this is the sort of information that one usually doesn't care very much about.

PR #189, opened 2017-08-16
Gfe integration tests

rowleya:
Change to make the stop function re-raise an exception if used to stop a binary that is running forever

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -2309,6 +2309,9 @@ class AbstractSpinnakerBase(SimulatorInterface):
         :rtype: None
         """
 
+        # Keep track of any exception to be re-raised
+        ex_type, ex_value, ex_traceback = None, None, None
+
         # If we have run forever, stop the binaries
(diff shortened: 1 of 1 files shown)


rowleya:
Part of the fix for SpiNNakerManchester/SpiNNakerGraphFrontEnd#58

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1846,9 +1846,13 @@ def _do_run(self, n_machine_time_steps, loading_done, run_until_complete):
...
-                self.stop(
-                    turn_off_machine=False, clear_routing_tables=False,
-                    clear_tags=False)
+                try:
+                    self.stop(
+                        turn_off_machine=False, clear_routing_tables=False,
+                        clear_tags=False)
+                except Exception:
+                    logger.error("Error when attempting to stop")
+                    traceback.print_exc()
```
dkfellows: Please write this to the log (at high-level priority) rather than just printing it out. Perhaps by combining with the previous line like this:
```py
logger.error("Error when attempting to stop", exc_info=True)
```

PR #191, opened 2017-08-18
Renaming key makefile

dkfellows:
It's always been deeply weird that we have makefiles with filenames that put `Makefile.` as a prefix. By far more conventional (and much better supported by IDEs) is to use `.mk` as a suffix. This branch changes that.

**NB:** This has a _substantial_ down-stream impact despite being a trivial change, as there's C code there that is built using this makefile. sPyNNaker is known to be impacted, though since the unification of the sPyNNaker implementation, this is much easier to handle. Also, we know that any users making their own models will have to make a (small, easy, but weird if not expected) change to accommodate this.


rowleya:
Due to the effects of this on other parts of the software, this needs to wait until the appropriate changes have been made elsewhere

dkfellows:
Nothing important left

PR #207, opened 2017-10-10
Fix spike injector database

rowleya:
Fixes SpiNNakerManchester/sPyNNaker#401


alan-stokes:
i know its not directly linked to the pull request. but that param your exploiting isnt documented in the constrctor python doc. Also, its set to false by default......

rowleya:
Is that better?

PR #210, opened 2017-10-12
Use 'install' to install, not 'cp'

dkfellows:
This switches the installation code in our makefile to use `install`, not `cp`. The `install` program gets more things right _for the installation case_ and makes it a little easier for people reading the code to see what is going on.


alan-stokes:
id rubber but think it a rowley world for that to verify.

PR #214, opened 2017-10-24
Multiple executable types per run

alan-stokes:
supports the ability to have multiple executableTypes running on the same application. Mainly used for reinjector and extra monitor support (which are to follow pull requests in future).

spynnaker and GFE depends upon this pull request
https://github.com/SpiNNakerManchester/sPyNNaker/pull/427
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/70


Christian-B:
On hold so that upstream repository branches can be given the same name and TESTED!

dkfellows:
As this is now passing tests, I'm minded to merge this. Once `master` is not broken…

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1966,8 +1969,8 @@ def _recover_from_error(self, exception, exc_info, executable_targets):
 
         # If there are any cores that are not in RTE, extract data from them
         if (len(non_rte_cores) > 0 and
-                self._executable_start_type ==
-                ExecutableStartType.USES_SIMULATION_INTERFACE):
+                ExecutableType.USES_SIMULATION_INTERFACE in
+                self._executable_types):
```
rowleya: It is worrying that nothing is below this change - you only want to extract data from those cores that are USES_SIMULATION_INTERFACE, otherwise the whole thing will grind to a halt.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1812,8 +1811,8 @@ def _do_run(self, n_machine_time_steps, loading_done, run_until_complete):
         if self._use_virtual_board:
             logger.warn(
                 "Application will not actually be run as on a virtual board")
-        elif self._executable_start_type == \
-                ExecutableStartType.NO_APPLICATION:
+        elif (ExecutableType.NO_APPLICATION in
+                self._executable_types):
```
rowleya: Presumably I can't have more than one executable type if there is NO_APPLICATION.  Perhaps it should be an error if NO_APPLICATION is there with other types...

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -79,12 +58,25 @@ def __call__(
...
-        # set off the executables that are in sync state
-        if sync_signal is not None:
-            txrx.send_signal(app_id, sync_signal)
-            txrx.wait_for_cores_to_be_in_state(
-                executable_targets.all_core_subsets, app_id,
-                [CPUState.RUNNING, CPUState.PAUSED, CPUState.FINISHED])
+        # set off the executables that are in sync state \
+        # (sending to all is just as safe)
+        if (ExecutableType.USES_SIMULATION_INTERFACE in
+                executable_types.keys() or
```
rowleya: Don't use .keys to check the keys of a dict, as this expands it to a list...

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -79,12 +58,25 @@ def __call__(
...
+                    executable_types, no_sync_changes)
+
+            # fire all signals as required
+            for sync_signal in sync_signals:
+                txrx.send_signal(app_id, sync_signal)
+
+        # verify all cores are in running states
+        txrx.wait_for_cores_to_be_in_state(
+            executable_targets.all_core_subsets, app_id,
+            [CPUState.RUNNING, CPUState.PAUSED, CPUState.FINISHED])
```
rowleya: Shouldn't this be determined by the executable types?

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -103,15 +95,43 @@ def __call__(
...
+        """ sorts out start states, and creates core subsets of the states for
+        further checks.
+
+        :param no_sync_changes: sync counter
+        :param executable_types: the types of executables
+        :return: the sync signal and updated no_sync_changes
+        """
+        sync_signals = list()
+
+        if ExecutableType.USES_SIMULATION_INTERFACE in executable_types.keys():
```
rowleya: Again on the use of .keys

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -103,15 +95,43 @@ def __call__(
...
+                sync_signals.append(Signal.SYNC0)
+            else:
+                sync_signals.append(Signal.SYNC1)
+            # when it falls out of the running, it'll be in a next sync \
+            # state, thus update needed
+            no_sync_changes += 1
+
+        # handle the sync states, but only send once if they work with \
+        # the simulation interface requirement
+        if ExecutableType.SYNC in executable_types.keys():
```
rowleya: Again...

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -452,3 +455,43 @@ def convert_time_diff_to_total_milliseconds(sample):
     :return: total milliseconds
     """
     return (sample.total_seconds() * 1000.0) + (sample.microseconds / 1000.0)
+
+
+def determine_flow_states(executable_types, no_sync_changes):
```
rowleya: It would make more sense to include this information in the ExecutableType enum data structure as ExecutableType.get_start_states and ExecutableType.get_end_states.  If these return a set, this function can then just return the union of the sets.

spinn_front_end_common/utility_models/live_packet_gather_machine_vertex.py
```diff
@@ -37,6 +37,8 @@ class LivePacketGatherMachineVertex(
                ('CONFIG', 1),
                ('PROVENANCE', 2)])
 
+    TRAFFIC_IDENTIFIER = "LPG_EVENT_STREAM"
```
rowleya: I don't quite see what this change has to do with this...

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
```diff
@@ -89,16 +88,21 @@ def test_mixed_binaries(self):
...
 
         graph = MachineGraph("Test")
         graph.add_vertices([vertex_1, vertex_2])
 
         gatherer = LocateExecutableStartType()
-        with self.assertRaises(ConfigurationException):
-            gatherer.__call__(graph)
+        results = gatherer.__call__(graph, placements=placements)
+        self.assertIn(ExecutableType.RUNNING, results)
+        self.assertIn(ExecutableType.SYNC, results)
```
rowleya: May also want to assert that these are the only things in there...

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
```diff
@@ -72,8 +71,8 @@ def test_call(self):
         targets = gatherer.__call__(
             placements, graph, _TestExecutableFinder())
         gatherer = LocateExecutableStartType()
-        start_type = gatherer.__call__(graph)
-        self.assertEqual(start_type, ExecutableStartType.RUNNING)
+        start_type = gatherer.__call__(graph, placements)
+        self.assertEqual(start_type.keys()[0], ExecutableType.RUNNING)
```
rowleya: Is this guaranteed to be 0 given it is a set?  Might be better checking size first...

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -875,8 +876,8 @@ def _run(self, run_time, run_until_complete=False):
                     break
 
         # Disable auto pause and resume if the binary can't do it
-        if (self._executable_start_type !=
-                ExecutableStartType.USES_SIMULATION_INTERFACE):
+        if (ExecutableType.USES_SIMULATION_INTERFACE not in
+                self._executable_types):
```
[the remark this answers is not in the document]
alan-stokes: you can. the other versions are not considered in the simulation calls. and for running, it is always in running state, so thats fine. sync is a bit intereting, as its meant to go to exit before auto pause and resume kicks in. so itll stay in exit state, as the runnign state is also allowing exit.

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: probabily true. will recheck

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: the extractor only extracts from cores using the simulation interface. the start type object has a map to cores. so this is safe to do

spinn_front_end_common/interface/abstract_spinnaker_base.py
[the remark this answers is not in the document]
rowleya: Hmm, not sure I like this then as those cores won't pause even though the simulation has paused.  I guess this depends on the semantics of those cores in that case...

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: OK, then this is fine :)

spinn_front_end_common/interface/abstract_spinnaker_base.py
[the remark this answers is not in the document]
alan-stokes: well in the cases we have. not pausing is exactly what we want them to do. extra monitor cores needs to be runnign for it to do the data speed up.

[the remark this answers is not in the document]
rowleya: Hmm yes, I can see why you need this here - still not sure I like it!  Basically you are saying that the only cores that are allowed to do this are those that we add.  In that case, maybe we should have a state that is something like ExecutableType.SYSTEM.  This would split it out from those cores that will reach exit state, or some intermediate sync state.

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: in this case, its output matching. the output is a dictonary. so it needs to chekc if its in the keys. it could also verify that length of keys is zero if thats what your going for. but i dont wnat to have multiple output tpyes, as itll complex the code even further

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: rubebred

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: to do that would require massive explosion of the types. you'd need a start, running, end flags separately. was not prepared to go down that road. This works as is. as all cores have to go into those states at some point

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: rubbered

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: we agreed to not go down this route.

spinn_front_end_common/interface/interface_functions/application_runner.py
rowleya: I think you already have the "end states" don't you?

spinn_front_end_common/utilities/helpful_functions.py
rowleya: But this means that if I add a type, I have to update this function...  That seems like an oversight...

spinn_front_end_common/utility_models/live_packet_gather_machine_vertex.py
alan-stokes: just a quick harmless cleanup. Given future changes will use this.

rowleya: Hmm OK !

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
alan-stokes: its a list, as you stated above about keys(). am testing that it detected the executable type correctly.

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
alan-stokes: rubebred

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
rowleya: I think it works in this case - keys() is a call into a dict and the returned list is of unspecified order.  I think that because this test expects only one thing, it is fine but just a warning...

rowleya: I suppose it should then check for size of 1 as well...

spinn_front_end_common/interface/abstract_spinnaker_base.py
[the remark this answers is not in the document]
alan-stokes: Im inclined to not do this until we find a use case. We could try anticipating the needs of invisible things. but unless theres a valid usecase, i think its wasted cycles.

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: if you show me how to build it, ill add it in. as i noted it was a better process, just didnt feel had time to do it.

unittests/interface/interface_functions/test_front_end_common_graph_binary_gatherer.py
alan-stokes: will add to please

alan-stokes: rubbered

spinn_front_end_common/interface/abstract_spinnaker_base.py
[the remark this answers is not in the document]
rowleya: I can see this going wrong quickly - again the sticky plaster solution.  We need to say what we mean - we already have two separate executable types here being specified as one (something that starts and ends vs. something that starts and never ends).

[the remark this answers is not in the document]
alan-stokes: fixed me thinks

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: fixed me thinks

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -781,9 +781,10 @@ def _run(self, run_time, run_until_complete=False):
         """
         self.verify_not_running()
         if (self._has_ran and
-                self._executable_start_type not in [
-                    ExecutableStartType.USES_SIMULATION_INTERFACE,
-                    ExecutableStartType.NO_APPLICATION]):
+                ExecutableType.USES_SIMULATION_INTERFACE not in
+                self._executable_types and
+                ExecutableType.NO_APPLICATION not in
+                self._executable_types):
```
rowleya: This could do to be updated - you can reuse supports auto pause and resume I think though...

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1812,8 +1812,8 @@ def _do_run(self, n_machine_time_steps, loading_done, run_until_complete):
         if self._use_virtual_board:
             logger.warn(
                 "Application will not actually be run as on a virtual board")
-        elif self._executable_start_type == \
-                ExecutableStartType.NO_APPLICATION:
+        elif (len(self._executable_types) == 0 and
+                ExecutableType.NO_APPLICATION in self._executable_types):
```
rowleya: These can't both be true...

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -79,12 +52,25 @@ def __call__(
...
+                    executable_types, no_sync_changes)
+
+            # fire all signals as required
+            for sync_signal in sync_signals:
+                txrx.send_signal(app_id, sync_signal)
+
+        # verify all cores are in running states
+        txrx.wait_for_cores_to_be_in_state(
+            executable_targets.all_core_subsets, app_id,
+            [CPUState.RUNNING, CPUState.PAUSED, CPUState.FINISHED])
```
rowleya: This should still be determined by the executable type I think - it can be done by unioning the set of end states can't it?

spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -103,15 +89,43 @@ def __call__(
...
+            # state, thus update needed
+            no_sync_changes += 1
+
+        # handle the sync states, but only send once if they work with \
+        # the simulation interface requirement
+        if ExecutableType.SYNC in executable_types:
+            if Signal.SYNC0 not in sync_signals:
+                sync_signals.append(Signal.SYNC0)
+
+        return sync_signals, no_sync_changes
```
rowleya: Doesn't this mean you can have multiple SYNC signals being sent - isn't that dangerous?  Maybe this can also be added to the ExecutableType

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: rubbered

spinn_front_end_common/utilities/utility_objs/executable_type.py
```diff
@@ -0,0 +1,62 @@
...
+    RUNNING = (
+        0,
+        [CPUState.RUNNING],
+        [CPUState.FINISHED],
+        False,
+        "Runs immediately without waiting for barrier and then exits")
+
+    SYNC = (
+        1,
+        [CPUState.SYNC0],
```
rowleya: I think this needs a FINISHED as well, in case it finishes before you look

spinn_front_end_common/utilities/utility_objs/executable_type.py
```diff
@@ -0,0 +1,62 @@
...
+from spinnman.model.enums import CPUState
+
+
+class ExecutableType(Enum):
+    """ supports starting of different types of executables
+    """
+
+    RUNNING = (
+        0,
+        [CPUState.RUNNING],
```
rowleya: This needs a finished as well...

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: doh. fixed. need to run these unit tets more often. lol

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: very true. fixing

alan-stokes: fixed

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: fixed

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: no way around it from what i can see. even putting it into the executable type has issue, as simulation changes its sync state based off reset.

spinn_front_end_common/utilities/utility_objs/executable_type.py
alan-stokes: yep, dead on. i put them in, then took them out, convincing myself it was safe. but nah ya right

spinn_front_end_common/utilities/utility_objs/executable_type.py
alan-stokes: yep, dead on. i put them in, then took them out, convincing myself it was safe. but nah ya right

spurred from donals comment of 1 state. lol

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: ya right.

spinn_front_end_common/interface/interface_functions/application_runner.py
rowleya: OK, so this state should never occur, because the SIMULATION binaries will go into SYNC0 in the first case anyway, thus all is well.  Maybe this should be documented though...

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: rubbered

spinn_front_end_common/utilities/utility_objs/executable_type.py
alan-stokes: rubbered

spinn_front_end_common/utilities/utility_objs/executable_type.py
alan-stokes: rubbered

spinn_front_end_common/interface/interface_functions/application_runner.py
alan-stokes: and its never multple of the same signal. jsut possible both sync0 and sync1

rowleya: So you should never send both sync0 and sync1 to the machine at the same time, and this code looks like it can, by adding sync1 from the SIMULATION code and SYNC0 from the SYNC code.  Thus if it ever gets to the point where it might, it should error.  That point should never occur in code from what I can see, but then who knows what someone might try in the future...

alan-stokes: why cant you? if they in different sync states, shouldnt sending both get them all to start? whats the limit here?

alan-stokes: fixed

PR #217, opened 2017-11-03
Separate multicast and fixed-route partition mappings

alan-stokes:
Allows multicast partitions and fixed-route partitions to be mapped separately and therefore not interfere with each other in key allocation.

This depends upon 
SPY https://github.com/SpiNNakerManchester/sPyNNaker/pull/424
GFE https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/69
PAC https://github.com/SpiNNakerManchester/PACMAN/pull/128


spinn_front_end_common/interface/interface_functions/edge_to_n_keys_mapper.py
```diff
@@ -125,28 +126,29 @@ def _process_application_partition(partition, n_keys_map, graph_mapper):
...
-        return constraints
+            for edge in partition.edges:
+                if isinstance(edge.post_vertex,
+                              AbstractProvidesIncomingPartitionConstraints):
+                    constraints.extend(
+                        edge.post_vertex.get_incoming_partition_constraints(
+                            partition))
+            constraints.extend(partition.constraints)
+
+            return constraints
```
dkfellows: This code fails to return an object (relying on Python's default) when `partition.traffic_type` is not `MULTICAST`. This feels like suspicious code-smell. If you want to return `None`, write so explicitly.

PR #223, opened 2017-11-14
Make non-optional algorithms non-optional

rowleya:
Fix current master issues!


rowleya:
The token set is that the data has been loaded - but neither algorithm can know that *all* the data has been loaded.  The token is thus nonsense.

PR #225, opened 2017-11-14
Add memory file interface

rowleya:
- [x] Add Smart Tokens to algorithms
- [x] Wait for merge of SpiNNakerManchester/PACMAN#135


Christian-B:
If multiple algorithms set the same token and the required order is that ALL of these algorithms must be run or skipped thee the Token can not ensure that.

So for multiple algorithms multiple tokens remain a requirement. But whatever decides to skip an algorithms must set a token.

rowleya:
No algorithms are now skipped... all are required as they should be

rowleya:
The token system can't be used to check that all these algorithms have run - all the next algorithm actually cares about is that one of them has been run :)

Christian-B:
If abstract_spinnaker_base.py selects to not add an algorithm it is skipped.  The issue remains that as order is token controlled there must be 1 token per possible algorithm otherwise order is not guaranteed.

Christian-B:
"all the next algorithm actually cares about is that one of them has been run :)"  So it is ok if run is executed before all the loads have finished??????

alan-stokes:
whats hes done is force all runs to use a data loading system during the load phase. this has now broken the ability to have a mapping algorithm do so instead. The rest of the software will handle it, as they track by their own special bools for if to run loading again anyhows. I still think lockign out a mapping data loader is a bad idea

dkfellows:
Can we have PACMAN work out the ordering of the phases, please?

rowleya:
Why would PACMAN do something I can do myself?

rowleya:
Note - this does *not* prevent a mapping algorithm from loading data and indicating it has done so.  Any mapping algorithm that loads data will likely do so on a vertex that doesn't support DSE or MemoryIO loading, or else if that vertex does so, it will *want* to run two phases of data loading.  Thus making these algorithms optional is incorrect.

rowleya:
This change allows an algorithm to do both DSE, MemoryIO loading and any other loading.

rowleya:
Note that all these algorithms produce an application_data_loaded_token.  Note that the algorithms are also both required algorithms.  Thus you have to run them at some point.  Now assume that I make an executor that orders things solely on the basis of the tokens, and still runs every algorithm I specify has to be run.  If that executor executes the binary loader *before* all algorithms that produce an application_data_loaded_token, it is generating a poor ordering.  I would suggest that such an executor should run *every* required algorithm that generates the token before running the thing that depends on it, otherwise it should be rewritten!

Christian-B:
"I would suggest that such an executor should run every required algorithm that generates the token before running the thing that depends on it, otherwise it should be rewritten!"    To the best of my knowlege Pacman does not support that.    Then what should it do if NO required algorithm generate the data but multiple optional ones do? Run 1, run all ?????

rowleya:
PACMAN supports this if the algorithms are in the correct order in the first place, otherwise it doesn't.  This pull request is not meant to fix PACMAN though, it works with the algorithm we have, and it is guaranteed to work with the algorithm we have, because the algorithms are ordered and all required.  If you provide two optional algorithms to PACMAN which produce the same output it will be guaranteed to only run one of them.  Hence why this pull request makes the algorithms non-optional.

rowleya:
OK, so I'm now realising there is a misunderstanding in what I want to represent here in terms of dependencies.  Tokens are fine in theory, but what I am trying to represent doesn't quite work with tokens.  The issue here is that the things that come after the data loaders need the data to have been loaded.  However they don't care how it was loaded.  Thus, and this is critical, these algorithms *don't* require the (currently) two algorithms to have run.  They don't care what has loaded the data, so long as someone has done it.  Therefore creating a token for each data loader is semantically incorrect.  In fact, in a simple token dependency system, this concept is simply not possible to represent without creating a direct link between the components (i.e. tokens for the specific algorithms).  The concept in this case is: 
 Every data loader algorithm that exists in the workflow, no matter how many, must have been run before the binary loader can run.

Note, also critically, it is not optional that those algorithms run.  They must be run, as we don't know at the outset which vertices have the property and which don't.  This should have also been the case for the DSE algorithm even in master, some vertices might load data during mapping and others might load data using DSE, and some might do both.  I have already verified that running this check is reasonable, even with 1,000,000 nodes (which takes around 0.3 seconds for an instance check; even with 10 million vertices this would take around 3 seconds per algorithm, and we are not going to have that many algorithms doing this!).

There are two ways to break this deadlock in PACMAN Executor:

- Accept that the ordering works with the tokens being used as a last check.  I can understand why you wouldn't want to accept this, as I know that a small change to one of the algorithm's inputs could result in a reordering of the execution, which could be incorrect.  For example, if the data loader suddenly needed an input from another algorithm, A, but the binary loaded didn't, it might end up running the binary loader first. Of course this could be avoided by just putting algorithm A earlier in the list; so long as the data loader doesn't depend upon an *output* of the binary loader, it should always be possible to make it run first.

- Change the PACMAN executor to encode the rule I mentioned above; when there is more than one algorithm that generates an output, O, it is an error unless *all* those algorithms can be run before one that requires O as an input.  I would imagine this could be quite hard to make work correctly, especially in light of algorithms that produce overlapping output.  However, if this is the only way to break the deadlock it could be done.  This would result in a fully token-based system, but there is no doubt that this increases the complexity of the code; given that people are already confused by it, I am not overly keen!

- Make the ordering completely fixed from the outset.  I don't think this works well, as then users can only add algorithms that run exactly at the start or the end, and it won't let their algorithm run somewhere in the middle.  Unless we stop users adding general algorithms, I don't think this is a good thing.  Thus I would say that only mapping would be customisable at that point, and that would have to use exactly the list provided in the config file.  This would have to include reports, as the way we currently do reports is to add them at the start which means they run as soon as they possibly can be.  If we did go down this road, I would probably suggest removing the executor from the other phases as it isn't really helpful - if you know the order you might as well just call the algorithms in that order and pass data as needed.

Any preferences or other ideas?

alan-stokes:
there's a 4th option here. We could have a intermediate process just for the loading phase, which takes all loading tokens and creates a final one which runner uses. this keeps the token system clean, satisfies your desire of the runner not caring how, but that its just done. And requires no changes to the pacman executor. It also means extra inptus etc wont be effected, as that becomes the point of confluence for all loaders. It means if we add a new one at a later date, the loader_grouper needs to be adjusted to take in that token as well. but i think thats a  fine compromise. 

this then allows us to remove the list behaviour, but i also ack that end users running the script twice having random order of algorithms (regardless of the fact they can be ran independently) could be a hard sell, so leaving the list order as  a nice to have for that is ok. 

Its worth noting that come future with parallel aspects, the order will be lost, and makes 3 almost pointless then. 

id keep the exeuctors in the other phases, as we dont know what the future will hold and it gives us that flex if we need it. removing the mapping, loading, data gen phases completely and having just one for all that and one for running would also remove the issues of optional vs required due to phases, as everything could become optional from FEC's view point, and the interfaces for front end / end user become the required ones. 

opinions?

rowleya:
There is no difference in having a token aggregation algorithm and having all the tokens needed by the final algorithm, except that it divides up the process.  My objection is based on the fact that if a user now adds a new algorithm, they will also have to override the token aggregation algorithm, which removes the user friendliness of the whole process.

The nature of the PACMAN Executor is that it is a token-based algorithm ordering system.  When there is more than one algorithm that can be run next (i.e. all the inputs are available for the next algorithm), it then picks based on the original ordering of the algorithms.  The argument here is then whether we should be allowed to take advantage of this arbitration order or not - I say why not, but others are more critical.  If we are allowed to take advantage of this order, we can do things as I have them now.  If not, I believe that the PACMAN Executor cannot meet the needs of this part of the execution, and therefore should be scrapped in place of simply calling the functions.  We can then just use the PACMAN Executor for running the mapping algorithms, since this is the only place that we expect users to add algorithms in any case, and there is less likely to be a multi-pre-token issue here (and if there is, we will probably have to make PACMAN a pure ordered execution algorithm, without doing an awful lot more work).

rowleya:
After yet more discussion, it was agreed that "Smart Tokens" are needed.  See issue SpiNNakerManchester/PACMAN#133.

alan-stokes:
comments!

PR #228, opened 2017-11-16
Minor changes: disentangle the code

dkfellows:
The code in `edge_to_n_keys_mapper.py` was needlessly hard to read. This sorts out:
* Fewer inverted tests
* Simpler code in conditionals
* No tests against variables with constant contents

It's intended to change no functionality, and just improve readability.


alan-stokes:
ya left a dangling code block 

 def _check_constraints_equal(constraints, stored_constraints):

i think that check is trying to verify if theres incompatibile constraints being added, but its not the way it should be done. to be hoenst, that code block looks wrong. so should likely be thrown out. it was never being ran in the first place.

alan-stokes:
after thats deleted, im happy to rubber

PR #232, opened 2017-11-21
Ripmcs detox

dkfellows:
This detoxifies the C code for the “reverse iptag multicast source” so that it stops using random bit bashing and instead just uses simple patterns via _vetted_ macros. This addresses a few bugs in the process where there were things wrong that we just happened to never hit in practice. I don't think it is a good idea for code to have such surprise “mines” in them, waiting to trip up random passing ships.

I guess quite a few of the macros (but not all) could be replaced by inlined functions. (The `UINT` macro can't work that way, but I suspect most of the others are convertible.)


c_common/models/reverse_iptag_multicast_source/reverse_iptag_multicast_source.c
```diff
@@ -605,63 +590,52 @@ static inline bool eieio_data_parse_packet(
...
+	if (pkt_payload_prefix > time) {
+	    add_eieio_packet_to_sdram(eieio_msg_ptr, length);
+	    return true;
+	}
+	late_packets++;
+	return false;
     }
 
-    if (pkt_type <= 1) {
+    if (!test_bit(pkt_type, EIEIO_PKT_32BIT)) {
```
rowleya: This again looks suspicious

alan-stokes:
leaving to rowley review

alan-stokes:
still leaving, as it seems has outstnading issues from rowley

dkfellows:
#462 effectively supersedes this

alan-stokes:
keeping branch for any reason??

dkfellows:
The parts not in #462 are even better handled in #613 so this can go totally now.

PR #235, opened 2017-11-23
Avoid load stuff not required after reset

Christian-B:
There was a considerable amount of stuff that was done after reset even when the application graph had not changed.

Interestingly enough one thing not done was to change the report folder.
Which meant some reports went pop when trying to recreate the same sub directory.

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1722,23 +1722,45 @@ class AbstractSpinnakerBase(SimulatorInterface):
                 self._config.get("Mapping", "loading_algorithms").split(","))
         algorithms.extend(self._extra_load_algorithms)
 
+        write_memory_report = self._config.getboolean(
+            "Reports", "write_memory_map_report")
+        if write_memory_report and application_graph_changed:
+            if self._exec_dse_on_host:
+                algorithms.append("MemoryMapOnHostReport")
+                algorithms.append("MemoryMapOnHostChipReport")
+            else:
+                algorithms.append("MemoryMapOnChipReport")
+
+        # Add reports that depend on compression
+        routing_tables_needed = False
+        if application_graph_changed:
+            if self._config.getboolean("Reports",
+                                       "write_routing_table_reports"):
+                routing_tables_needed = True
+                algorithms.append("unCompressedRoutingTableReports")
+                algorithms.append("compressedRoutingTableReports")
+                algorithms.append("comparisonOfRoutingTablesReport")
+            if self._config.getboolean(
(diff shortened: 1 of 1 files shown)


alan-stokes:
request to remove self._application_graph_changed from a class variable to just a local. the risks of keeping it updated are more dangerous as class variable

alan-stokes:
1705 merge the ifs

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1677,20 +1677,34 @@ def _do_load(self):
...
         # add report for extracting routing table from machine report if needed
         # Add algorithm to clear routing tables and set up routing
-        if not self._use_virtual_board:
-            algorithms.append("RoutingSetup")
-            # Get the executable targets
-            algorithms.append("GraphBinaryGatherer")
+        if not self._use_virtual_board and application_graph_changed:
+                algorithms.append("RoutingSetup")
+                # Get the executable targets
+                algorithms.append("GraphBinaryGatherer")
```
dkfellows: Bonus indentation? Will flake8 like it?

alan-stokes: anwser = no

alan-stokes: does flake8 like anything

dkfellows: Meh. It's allowed, but it still sucks slightly.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1677,20 +1677,34 @@ def _do_load(self):
...
             algorithms.append("GraphBinaryGatherer")
 
-        if helpful_functions.read_config(
-                self._config, "Mapping", "loading_algorithms") is not None:
+        loading_algorithm = helpful_functions.read_config(
+            self._config, "Mapping", "loading_algorithms")
+        if loading_algorithm is not None and application_graph_changed:
             algorithms.extend(
                 self._config.get("Mapping", "loading_algorithms").split(","))
         algorithms.extend(self._extra_load_algorithms)
```
dkfellows: Why are we not using the variable that we created a few lines above?

dkfellows: `loading_algorithm` that is…

dkfellows: We could just do `algorithms.extend(loading_algorithm.split(","))` and all would be good.

alan-stokes: good catch. i didnt notice when i did the refactor

PR #245, opened 2017-12-07
Provide Pre Allocs

rowleya:
This fixes master when using spalloc by providing the missing optional token, which allows some algorithms to run earlier than they otherwise would.


rowleya:
Better than this is to instead pull SpiNNakerManchester/PACMAN#140.  This solves the problem in a more consistent way.

rowleya:
If you decide the pull the PACMAN change, this one can be deleted, as it isn't needed in that case.

rowleya:
Closing as decided not a good idea

PR #256, opened 2017-12-14
2 graph multi run reset fix

alan-stokes:
allows to adapt to spalloc machines which come with different formats as the original one it was mapped to. 

has a spynnaker related one as well. 
https://github.com/SpiNNakerManchester/sPyNNaker/pull/450


rowleya:
A minor comment but not a blocker (i.e. if you want to change this, create a new branch) -the machine graph is created regardless of it there is an application graph already, but then gets recreated if there is an application graph.  It doesn't break anything, and doesn't cost much, but just a comment to note it...

PR #263, opened 2017-12-21
issue with logic flow

alan-stokes:
seems finisher never got tested, and never fails unless you have mixed executable types. this fix fixes it though.

tested with robert code

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/application_finisher.py
+++ b/spinn_front_end_common/interface/interface_functions/application_finisher.py
@@ -25,8 +25,9 @@ class ApplicationFinisher(object):
             "Turning off all the cores within the simulation")
 
         # check that the right number of processors are finished
-        processors_finished = txrx.get_core_state_count(
-            app_id, ExecutableType.USES_SIMULATION_INTERFACE.end_state[0])
+        processors_finished = sum(
+            txrx.get_core_state_count(app_id, state)
+            for state in ExecutableType.USES_SIMULATION_INTERFACE.end_state)
         finished_cores = processors_finished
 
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/interface_functions/application_finisher.py
```diff
@@ -26,7 +26,7 @@ def __call__(self, app_id, txrx, executable_types):
 
         # check that the right number of processors are finished
         processors_finished = txrx.get_core_state_count(
-            app_id, CPUState.FINISHED)
+            app_id, ExecutableType.USES_SIMULATION_INTERFACE.end_state[0])
```
rowleya: Why only 0?  Should it add up all possible end_states?

alan-stokes: well there's only one finished state for simulation interface. you could put a iterator on there, and add them all up and aggregate the total. But seemed a bit over kill (plus i was trying to figure out why the thing was blowing up in the first place. lol. get_core_state_count doesn't like lists it seems. lol

rowleya: I think the option is either to do the iteration or make the transceiver version accept a list.  The issue is protection against future changes - if we change this so it has more than one state, this bit will go wrong.

The simple solution is to use the python sum construct:
```
processors_finished = sum(
    txrx.get_core_state_count(app_id, state) 
    for state in ExecutableType.USES_SIMULATION_INTERFACE.end_state)
```

alan-stokes: fixed as requested

PR #270, opened 2018-01-22
Structural plasticity

pabogdan:
Deleting some data files so it doesn't fill disk


dkfellows:
I'm minded to accept this

dkfellows:
If we do not ever want to keep the `reloaded_data_regions` directories, this is an entirely correct change. I do not know if we want to be able to keep the information around for debugging purposes.

pabogdan:
@rowleya , any thoughts?

rowleya:
I don't think we can remove all of the loaded data files; I think we need the last one but I need to look at the code to confirm (it might be that we regenerate the last one anyway).  The failure case to check for would be run, change parameter, run, reset, run; after reset we load the data from the initial generation, and then load on top any changes to parameters.

dkfellows:
Brainstorming note: possible config parameter name: `keep-dsg-reload-data-files`. @Christian-B suggests that in full debug mode, that should be definitely on, and I suggest that that should be off normally.

But need consideration of what to delete and what to retain in normal mode; I don't have a good feel for that at all.

dkfellows:
That's the change that's needed except for the code for the decision of whether to delete.

I also factored out the “make a directory name” and “create the directory” code because they're _hugely_ better when fused to each other more closely.

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
```diff
@@ -86,6 +81,20 @@ def __call__(
         for vertex in application_vertices_to_reset:
             vertex.mark_regions_reloaded()
 
+        # FIXME: Needs correct logic for whether to delete
+        if self._DELETE_TEMP_DIRS:
```
dkfellows: This is the line that needs to be something else, but I don't know what.

rowleya:
It looks like we *can* currently remove all the data files, but this is because the run, change parameter, run, reset, run cycle is currently broken (see #298).

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
alan-stokes: having reviewed, i also think a cfg param so that we can change it without going into code.

dkfellows:
I'll commit the change to `data_specification.c` to master; it's just changing a `log_info` into `log_debug` and the delay on the other part is unrelated.

Christian-B:
Oops. Never should have been deleted or closed. Sorry Peter!

alan-stokes:
Is this brnach staying for any real reason

pabogdan:
nope

PR #273, opened 2018-01-26
New speed up fr protocol

alan-stokes:
this branch supports a more optimised Fixed route protocol to double the output (i hope) from each board.

issue currently is that the C code cannot keep up with it, and therefore is resulting in reinjection mode, and in the end, results in slower performances than not using. 

this is left as a PR on the hope that the C code can be improved to a point where using the C code is particle. 

this PR also allows the end user to configure if they should use the python code vs C code. this would be useful as a standalone, so a cherry pick may be worth while.

- [ ] ensure that the C code can support this bandwidth
- [ ] verify that the python can complete without error.


Christian-B:
please resolved conflicts so I can review/test

alan-stokes:
circumvented by parallel c++ pr. closing

PR #274, opened 2018-02-05
adds speed up report

alan-stokes:
used for debug when other cores go boom in unsetting the timeouts. this report will be allowing us to determine if the speed up data flowed through the chip.

includes a Spynnaker8 for testing this:
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/111


alan-stokes:
good question......... dont see the harm in putting a config param in (par the mounting pile of them). either your welcoem to add it, or itll need to wait till ive finished having my arms deep in data in guts.

rowleya:
Happy enough with that now...

PR #277, opened 2018-02-12
Set emergency timeouts

rowleya:
This fixes the settings of the emergency router timeouts to avoid a hardware bug


spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
```diff
@@ -272,22 +278,58 @@ def get_local_provenance_data(self):
...
+                    extra_monitor_cores_for_router_timeout)
+        except Exception:
+            logger.error("Error resetting timeouts", exc_info=True)
+            logger.error("Checking if the cores are OK...")
+            core_subsets = convert_vertices_to_core_subset(
+                extra_monitor_cores_for_router_timeout, placements)
+            try:
+                error_cores = transceiver.get_cores_not_in_state(
+                    core_subsets, {CPUState.RUNNING})
+                if len(error_cores) > 0:
```
dkfellows: This probably ought to be `if error_cores:`, since it's known to be a Python list.

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
```diff
@@ -272,22 +279,57 @@ def get_local_provenance_data(self):
                                     n_lost_seq_nums))))
         return prov_items
 
-    @staticmethod
     def set_cores_for_data_extraction(
-            transceiver, extra_monitor_cores_for_router_timeout,
+            self, transceiver, extra_monitor_cores_for_router_timeout,
```
dkfellows: Can't we call this parameter `extra_monitor_cores`? The extra length doesn't add very much information yet it forces a lot more backslashes in the code.

Also, this is a public method without any documentation comment.

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
```diff
@@ -272,22 +279,57 @@ def get_local_provenance_data(self):
...
             extra_monitor_cores_for_router_timeout)
+        extra_monitor_cores_for_router_timeout[0].\
+            set_reinjection_router_emergency_timeout(
+                1, 1, transceiver, placements,
+                extra_monitor_cores_for_router_timeout)
 
-    @staticmethod
     def unset_cores_for_data_extraction(
-            transceiver, extra_monitor_cores_for_router_timeout,
+            self, transceiver, extra_monitor_cores_for_router_timeout,
```
dkfellows: Same problems; excessively long name and no documentation comment on a public method

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
rowleya: Happy enough to make these changes, but these aren't changes made here (the problem you mention is already in master).

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
rowleya: As above...

alan-stokes:
looks fine to me. will need to retune code when this goes in for speed in, as been using that function quite a bit. but thats just really a note to self

PR #280, opened 2018-02-16
Pop views assembles

dkfellows:
The main changes here are nearly 2 years old… so bringing them up to date is going to be “challenging”…


alan-stokes:
closing as obviously this work is now a dead end work.

alan-stokes:
christian can delete branch when he feels hes used whatever nuggets are in there.

Issue #283, opened 2018-02-20
Database is written in every run


georghinkel:
If you repeatedly call run, e.g. by repeatedly calling run through the PyNN API, the DatabaseInterface algorithm is always executed which in turn always writes the database. 

For applications that call run repeatedly (such as the Neurorobotics Platform developed in HBP SP10), this causes a lot of overhead for times between run calls. For a small test network, it took about 3s where we are calling run with about 100ms as time argument. It would be great to write the database only whenever this is really necessary in order to reduce the overhead and therefore improve the ratio between actual simulation time and simulation time plus overhead between repeated run calls.

alan-stokes:
hopefully fixed in 
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/284

andrewgait:
Looks to me like this has been fixed; please open again if it still needs work.

PR #284, opened 2018-02-20
only write database on first run and remapping

alan-stokes:
i think this works, but cant test. laptop gone weird and turned into read only on the partition with software


alan-stokes:
hopefully fixes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/283

rowleya:
This makes loads of changes that seem irrelevant to this change and it fails loads of tests.  Is there something specific this is supposed to be changing?

alan-stokes:
unknown why thers that many line changed. it should have only 4 or 5 lines changed. ...... maybe due to being so far out from master?

rowleya:
The state you see here is at the point when it was created.  It looks like this was started from an unclean point...  It may be worth redoing this if the change is easy enough.  I can't work out which part was meant to be changed or I would do it myself!

alan-stokes:
added to my pile for after data in fixed.

andrewgait:
> The state you see here is at the point when it was created. It looks like this was started from an unclean point... It may be worth redoing this if the change is easy enough. I can't work out which part was meant to be changed or I would do it myself!

I agree with this - this probably needs redoing from a more up-to-date point in master.

dkfellows:
Data In should now be fixed enough that this can be reconsidered.

rowleya:
That's a bit simpler!

rowleya:
I believe that this now works having tested it (it writes the database at the start if needed and then only if the graph changes)

alan-stokes:
thanks for picking this up.

Issue #288, opened 2018-02-27
Value error on truth values when running fixed number connectors


andrewgait:
Running e.g. https://github.com/SpiNNakerManchester/sPyNNaker8/blob/master/p8_integration_tests/test_fixed_number_connectors/test_fixed_number_pre_connector.py gives the following:

===============================================2018-02-27 11:27:01 ERROR: Error when calling pacman.operations.partition_algorithms.partition_and_place_partitioner.PartitionAndPlacePartitioner.__call__ with inputs {'machine': [Machine: max_x=1, max_y=1, n_chips=4], 'graph': <pacman.model.graphs.application.application_graph.ApplicationGraph object at 0x7f48bd676848>, 'preallocated_resources': <pacman.model.resources.pre_allocated_resource_container.PreAllocatedResourceContainer object at 0x7f48babee730>}
Traceback (most recent call last):
  File "/localhome/mbbssag3/spinnaker/git/sPyNNaker8/p8_integration_tests/test_fixed_number_connectors/test_fixed_number_pre_connector.py", line 197, in <module>
    stim_spikes, spklist_exc, spklist_inh = do_run(plot=True)
  File "/localhome/mbbssag3/spinnaker/git/sPyNNaker8/p8_integration_tests/test_fixed_number_connectors/test_fixed_number_pre_connector.py", line 141, in do_run
    p.run(sim_duration)
  File "/localhome/mbbssag3/spinnaker/git/sPyNNaker8/spynnaker8/__init__.py", line 528, in run
    return __pynn_run(simtime, callbacks=callbacks)
  File "/localhome/mbbssag3/Downloads/PyNN/PyNN-0.9.1/PyNN9/pyNN/common/control.py", line 111, in run
    return run_until(simulator.state.t + simtime, callbacks)
  File "/localhome/mbbssag3/Downloads/PyNN/PyNN-0.9.1/PyNN9/pyNN/common/control.py", line 93, in run_until
    simulator.state.run_until(time_point)
  File "/localhome/mbbssag3/spinnaker/git/sPyNNaker8/spynnaker8/spinnaker.py", line 127, in run_until
    self._run(tstop - self.t)
  File "/localhome/mbbssag3/spinnaker/git/sPyNNaker8/spynnaker8/spinnaker.py", line 174, in _run
    AbstractSpiNNakerCommon.run(self, duration_ms)
[...]

PR #289, opened 2018-02-27
check if send_buffer_times exists, truth of array is ambiguous

andrewgait:
Fix for issue https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/288 which happened when I updated to recent master and tried to run fixed-number connectors; I don't know precisely why it's only happening with fixed number connectors and nothing else (as far as I can tell), but this change seems to fix the problem.

### DIFF
--- a/spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
+++ b/spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
@@ -224,7 +224,8 @@ class ReverseIPTagMulticastSourceMachineVertex(
             self._install_virtual_key(n_keys)
 
     def _install_send_buffer(self, n_keys, send_buffer_times, target_address):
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
```diff
@@ -224,7 +224,8 @@ def __init__(
             self._install_virtual_key(n_keys)
 
     def _install_send_buffer(self, n_keys, send_buffer_times, target_address):
-        if send_buffer_times and hasattr(send_buffer_times[0], "__len__"):
+        if (send_buffer_times is not None and
```
rowleya: This must have slipped in from a previous linting.  It should be if len(send_buffer_times) and hasattr... as this is checking whether send_buffer_times has some values in it.  A prior check already checks for None.

PR #292, opened 2018-03-02
Spalloc fixes

### LINKED ISSUE #291: spalloc ignoring cfg settings like max_sdram_allowed_per_chip
Found with 
https://github.com/SpiNNakerManchester/sPyNNaker8/blob/master/p8_integration_tests/test_auto_pause_and_resume_tests/test_synfire_1_run_no_extraction_if_curr_exp_low_sdram.py

exrta monitor support on.

as @alan-stokes noted in skype:
"yep, if you look at xml for itnerface functions. spalloc and hbp alooc dont take that as a input, they generate it as a outputso they basically ignore this parameter from the cfg. #284 

even more, they return None for that value, so shouldnt be hard to push it in"

The above test will need to be changed to expect a failure.
add to config file (to make sure these are used)
[machine]
enable_advanced_monitor_support = True
enable_reinjection = True

Original test will then need to be copied in another directory with a higher 
[Machine]
max_sdram_allowed_per_chip = 3000000

rowleya:
Fixes #291 

Also adds some calls to the HBP allocation that will make the interaction work better.  Although this requires the HBP portal to also be updated (see SpiNNakerManchester/RemoteSpiNNaker#23), this can still be merged now, as this software would also have to be released on the portal VM before it became an issue.

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/front_end_common_interface_functions.xml
+++ b/spinn_front_end_common/interface/interface_functions/front_end_common_interface_functions.xml
@@ -805,6 +805,10 @@
                 <param_name>down_links</param_name>
                 <param_type>DownedLinksDetails</param_type>
             </parameter>
+            <parameter>
+                <param_name>max_sdram_size</param_name>
+                <param_type>MaxSDRAMSize</param_type>
+            </parameter>
         </input_definitions>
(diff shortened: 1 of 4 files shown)


rowleya:
That should fix it...

PR #294, opened 2018-03-06
Fix buffered out

rowleya:
Another fix for buffered out.  Errors this time were:
 - the reads back from the system weren't synchronised, so a read could be in progress when get_data_for_vertex was called
 - the reads during simulation attempted to use the faster data extraction, which is known not to work during simulation due to a need to set the router timeouts to infinite

This fixes these issues.

### DIFF
--- a/spinn_front_end_common/interface/buffer_management/buffer_manager.py
+++ b/spinn_front_end_common/interface/buffer_management/buffer_manager.py
@@ -603,7 +603,8 @@ class BufferManager(object):
 
         # Ensure that any transfers in progress are complete first
         with self._thread_lock_buffer_out:
-            self._get_data_for_vertex_locked(placement, recording_region_id)
+            return self._get_data_for_vertex_locked(
+                placement, recording_region_id)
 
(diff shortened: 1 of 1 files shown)


alan-stokes:
looks good, par two little things. 
1. @property
def reload_buffer_files(self):
 return self._reload_buffer_file_paths

that variable don't exist.
I get its not part of the change, but i noticed it during review.

2.  ya changed the interface for recording_utilities. but i don't see a spynnaker pr for those changes.

rowleya:
It is worth noting that this probably isn't up to date with master, but equally doesn't conflict with master.  If you did your comparison using PyCharm, you will see changes that aren't actually there.  The things to review are really those in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/294/files (i.e. the Files changed link above).

rowleya:
To be clear, the *only* changes in this branch are to buffer_manager.py.

alan-stokes:
merged master in. and indeed those api changes disappear. but the missing variable still stands

PR #301, opened 2018-03-22
Speed up data in optimsied dma and sdram

alan-stokes:
adds the SDRAM removal to DTCM, and utilises DMA's in the data receiver. Alas it doesn't gain any improvements....... so might not be worth putting in. but meh. 

This also seems to be a bit flaky in spalloc............ odd


dkfellows:
Also, with the new codebase it will require some way of doing the DMAs in the gatherer as well, as that doesn't use the extra monitor on the ethernet chip.

dkfellows:
See the `data-in-dma` branch for a newer version of this

PR #302, opened 2018-03-22
Speed up optimised sdram to dtcm

alan-stokes:
this removes sdram requirement from the extra monitor core. so gets us approx 12k sdram back to the application


dkfellows:
Changes concerned will be considered as part of #301

PR #303, opened 2018-03-23
Parallel c++

alan-stokes:
supports faster data out.

- [x] works
- [ ] is releasable

depends upon
 - [ ] support scripts:  https://github.com/SpiNNakerManchester/SupportScripts/pull/10
 - [ ] spinnaker tools: https://github.com/SpiNNakerManchester/spinnaker_tools/pull/72


cpp_common/common/UDPConnection.cpp
```diff
@@ -0,0 +1,227 @@
...
+	local_address = get_address(local_host, local_port);
+    } else {
+	sockaddr_in *local = (sockaddr_in *) &local_address;
+	local->sin_family = AF_INET;
+	local->sin_addr.s_addr = htonl(INADDR_ANY);
+	local->sin_port = htons(local_port);
+    }
+
+    // TODO: What is the correct timeout for a receiver socket?
+    setSocketTimeout(sock, 500ms);
```
dkfellows: So… what _is_ the correct timeout for a socket here? (Timeouts just cause the reader loop to go round again, which allows us to detect termination cleanly and costs next to nothing.)

dkfellows:
No longer relevant; obsoleted by #372

PR #304, opened 2018-03-23
Data in parallel sdram to dtcm

alan-stokes:
work on getting c code parallisation

is a build onto of all rest of data in work


dkfellows:
No longer relevant, and too much divergence

PR #306, opened 2018-03-28
Add reset of reloadable regions after DSG

rowleya:
This is needed to avoid a call which changes something meaning that DSG reload occurs when it shouldn't.  If you have done DSG, we can be sure the region has been reloaded after all...


alan-stokes:
might be getting confused, but if youve changed, then reset and ran, wont the flag have been set to not reload at that point?

rowleya:
Note that this has now changed to reflect required changes to sPyNNaker, and to avoid having to replicate those changes.  See SpiNNakerManchester/sPyNNaker#498 for details.

oliverrhodes:
This fixes the issue I saw with the Poisson Spike source: data is no longer re-written after a run, when before the run the rate was changed.

@alan-stokes please merge when you've reviewed

PR #308, opened 2018-04-12
Updated to allow users of the makefile to select the optimisation

rowleya:
The default optimisation is left at $(OTIME) but this at least now lets the user override it


Christian-B:
I am in the process of completely reworking what was currently FrontEndCommon.mk

So this PR would be lost if approved.

PR #311, opened 2018-04-16
Fix spalloc close

rowleya:
Avoids closing spalloc client twice

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
+++ b/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
@@ -59,6 +59,7 @@ class _SpallocJobController(MachineAllocationController):
     def _teardown(self):
         if not self._exited:
             self._job.close()
+        super(_SpallocJobController, self)._teardown()
 
 
 class SpallocAllocator(object):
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
```diff
@@ -57,7 +57,8 @@ def _wait(self):
 
     @overrides(MachineAllocationController._teardown)
     def _teardown(self):
-        self._job.close()
+        if not self._exited:
```
dkfellows: Needs to call superclass implementation of this method. _After_ this test/`if`.

dkfellows: NB: the method was wrong before. Doesn't mean it should stay wrong.

rowleya: I guess it should - teardown is a "pass" method just now, but it isn't abstract so it doesn't have to stay that way...

rowleya: Done - this happens regardless of whether the if is met (which I think is correct).

dkfellows: It only mattered because that super call is how the `_exited` field gets set.

PR #315, opened 2018-04-26
Dictionary for log messages

Christian-B:
https://github.com/SpiNNakerManchester/spinnaker_tools/pull/75

FrontEndCommon.mk split into local.mk and neural_support.mk (other repositories)

Most other makefile changes are renaming of common


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2690,7 +2686,7 @@ def _extract_iobufs(self):
         extractor = ChipIOBufExtractor()
         extractor(
             transceiver=self._txrx,
-            core_subsets=self._last_run_outputs["CoresToExtractIOBufFrom"],
+            executable_targets=self._last_run_outputs["ExecutableTargets"],
```
alan-stokes: this has resulted in the loss of the iobuf control flags in cfg. will need to be reengineered

c_common/front_end_common_lib/local.mk
```diff
@@ -62,13 +62,13 @@ all: $(ALL_TARGETS)
 
 # All the c and h files built at the same time but individual rules needed for make chains
```
alan-stokes: think this file should be renamed to something akin to FrontEndCommon.mk

c_common/models/fec_models.mk
```diff
@@ -0,0 +1,7 @@
+MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
```
alan-stokes: by changing the order of how the main make here works, you could remove this 
+include $(CURRENT_DIR)../front_end_common_lib/local.mk
to point at the spinnaker_tools version.
aka.a build lib, install lib, build models, install models.

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: fixed!

c_common/front_end_common_lib/local.mk
Christian-B: Used front_end_common.mk to keep it distinct from the now defunct  FrontEndCommon.mk

c_common/models/fec_models.mk
Christian-B: Done

rowleya:
I can see that this makes the FEC Makefile less complex, but this seems to be at the cost of the complexity of the neural and extra models Makefiles.  I can see some commonalities, so I am going to have a look and see if these can be generalised in some way, without increasing the complexity of the FEC makefile again!

Christian-B:
Very open to suggestions but hopefully without the previous double define stuff.

rowleya:
I believe that my refactoring of this is done - comments welcome!

PR #320, opened 2018-05-03
allowing sub partitioning of keys for allowing retina/ssa to be filtered via routing

alan-stokes:
depends upon:
 PACMAN  https://github.com/SpiNNakerManchester/PACMAN/pull/158
SPY https://github.com/SpiNNakerManchester/sPyNNaker/pull/526

- [x] make work
- [ ] reviewed


rowleya:
I don't think this does what it says it does i.e. it doesn't allow a retina to be used with only part of the retina.  From what I can see all that these branches add is a check for a specific case where the thing being targeted has an Incoming Key Constraint (which pretty much nothing in our tools have that I know of).  The check makes sure that the incoming key constraint covers all of the keys of the sender, assuming that the sender uses a fixed key.  If I pick only the "up" spikes from a retina, and give it the key and mask for just this, I can't see that this will fail...

Christian-B:
Too far behind and idea now done better elsewhere

Issue #322, opened 2018-05-03
AbstractSpinnakerBase._get_machine ignores n_chips_required


Christian-B:
AbstractSpinnakerBase._get_machine
when hostname is None So not spalloc
calls:
GetMachineProcess which does not have a notion of n_chips_required

then NO checking is done to make sure the returned machine is big enough!

Christian-B:
fixed

PR #323, opened 2018-05-04
Fail fast if n_chips_required not possible

Christian-B:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/322


alan-stokes:
ive got a niggle, but im willing to accept its a tiny niggle and ignorable. 

spalloc checks the chips required before hanging back, via a error if memory serves. 
couldn't virtual and real do the check internally, instead of it being checked after the entire flow has been ran for get_machine()?

but by pushing the check into those functions, i ack that there'd be duplicate code, or a common call in utilies which may be less clean than where it is now........

PR #331, opened 2018-06-15
Allow exception in debug

rowleya:
Allows the use of debuggers when using the tools


alan-stokes:
"Allows the use of debuggers when using the tools" in what sense is sending a error messages to  sys.stderr supporting the use of the debugger?

rowleya:
In what sense does this only send error messages to sys.stderr?

So, to explain in more detail, a debugger normally (as far as I know - certainly in Eclipse) registers itself as an exception handler, overriding the default behaviour.  Our tools do exactly the same, but prior to this pull request, they ignored the old exception handler.  This pull request changes this by calling the original exception handler, which should trigger the debugger.

alan-stokes:
this came from https://docs.python.org/2/library/sys.html

sys.excepthook(type, value, traceback)
This function prints out a given traceback and exception to sys.stderr.

thats why my question came through. But your info means i can see the reasoning now.

PR #335, opened 2018-07-13
Fix 2 issues with memory calculations.

rowleya:
1) Only the memory for recorded items was accounted for
2) The minimum was calculated per chip so only the last actually mattered


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1100,29 +1100,32 @@ def _deduce_number_of_iterations(self, n_machine_time_steps):
...
-            min_time_steps = min(
+            min_time_steps_this_chip = min(
                 int(vertex.get_n_timesteps_in_buffer_space(
                     sdram_per_vertex, self._machine_time_step))
                 for vertex in vertices_on_chip)
+            if min_time_steps is None:
+                min_time_steps = min_time_steps_this_chip
+            else:
+                min_time_steps = min(
+                    (min_time_steps, min_time_steps_this_chip))
```
dkfellows:
```suggestion
                    min_time_steps, min_time_steps_this_chip)
```

PR #340, opened 2018-07-25
supports spinnaker_2 vm

alan-stokes:
still works with our normal setups. but allows a vm which reflects the cores and routers of a spinnaker2


spinn_front_end_common/interface/spinnaker.cfg
```diff
@@ -41,7 +41,7 @@ critical =
 reports_enabled = True
 write_energy_report = False
 write_text_specs = False
-write_router_reports = False
+write_router_reports = True
```
rowleya: If I remember correctly, this makes things very slow on big networks...

spinn_front_end_common/interface/spinnaker.cfg
```diff
@@ -106,6 +106,8 @@ spalloc_machine = None
 
 virtual_board = False
 requires_wrap_arounds = None
+NCoresPerChip = 18
```
rowleya: Would it be better to make this 16 by default since this is what it was in the current code?

spinn_front_end_common/interface/spinnaker.cfg
```diff
@@ -106,6 +106,8 @@ spalloc_machine = None
...
+RouterTableEntriesPerRouter = 1024
```
rowleya: Should this be 1023 to work with "normal" circumstances i.e. one entry taken by SCAMP?

alan-stokes: well i changed it, but the default router says there's 1024 entries. as line 17 of router states. So we're always routing with 1024 currently. but changed anyhow.

spinn_front_end_common/interface/spinnaker.cfg
alan-stokes: but i takes out one core to monitor in spinnMachine. so yourd really argue 17, but i dont know. ive put it as 16 anyhow.

spinn_front_end_common/interface/spinnaker.cfg
alan-stokes: moved

PR #341, opened 2018-08-09
Allows vm reporting of cores and links

alan-stokes:
alas depends on max machine, coz i didnt relise where i was. but this allows the vm to report its cores and links. which is the expected behaviour of our tools when buidling a machine. But only happens when a spinnMan is invovled currently.


Christian-B:
This branch has conflicts if I merge in master on VirtualMachineGenerator

alan-stokes:
i boiled this down to what the pr was on about. the stuff in ASB i think was trying to separate max machine from machine. but given i cant remember why, and that it was so far behind and had such a massive conflict in the key area. i just killed it. if the bug ever returns we can fix it then and there. 

so this is now a trival change and should be accepted with no problem now

rowleya:
This is still behind master... maybe why Travis isn't working on it?

alan-stokes:
bought back up with master. is travis now working?

PR #347, opened 2018-09-07
adds citation and manifest files

alan-stokes:
adds citation and manifest files. start point for further work on creating a doi for the tools


alan-stokes:
issue is related to being linked to support scripts. any opinion of either where to move the code to. or if we should release support scripts?

rowleya:
I don't think we should release support scripts; there should be no need as the citation stuff is nothing to do with running the code itself.  Instead, the citation stuff should be optional to run...

alan-stokes:
the reason its not optional is that the end user might not know which front end they actually used. Thus why i'm operating the aggregator code from fec stop (https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/347/files#diff-818c33345a4db6fa3fe6679f333f2389R2628)

otherwise the code is too hidden for any user to ever use. 
I guess i could put the aggregator code in SpiNNUtils to avoid the support scripts issue.    

im not going to run the doi generator during each run, the end user will have to do that. But adding more walls for the end user to do seems more likely to stop them from using it.

rowleya:
I think that doing the citation aggregation and warning the user each time the user uses the software might be a bit overkill!  We can add to or edit the license terms but that is about all we should be doing...

Having the Aggregator in SpiNNUtils is OK so long as it doesn't have too many additional requirements.

rowleya:
Maybe this can be an additional script rather than built in to the tools...

alan-stokes:
pooy. k. will remove from fec. 

i would have thought having it generate each time, and then supporting a p.generate_bibtex() would have been a good idea, as everything is in the end users field. but ok.

rowleya:
p.generate_bibtex is fine... couldn't that do the actual aggregation at that point?

rowleya:
Happy if it doesn't mean SpiNNUtils depends on lots of things!

alan-stokes:
the only two i think are not built in packages are yaml and requests. io might not be, but google isn't being helpful there. 

yaml is for reading and writing the CITATION.cff as thats the foramt they're using,
requests is the rest api for Zenodo.

both are pip install from my understanding. lets put it this way, i dont even renember installing them

alan-stokes:
ok, theres a 3rd. 
FEC will need to depend on cffconvert for this to work now. its a pip install so thats at least hidden

rowleya:
OK, that is fine

PR #348, opened 2018-09-28
Warn on startup

rowleya:
Warn rather than fail if the application has not been detected to start


rowleya:
Looking at this, I think it might be better to just not do this check at all.  At this point, all the cores are guaranteed to have started.  The check done here is to check that the sync start has worked.  This is fine in theory but in practice is bad for a couple of reasons:
 1. The check sends additional packets in to the network, and so could affect the running of the network.
 1. The check takes a bit of time, after which the notification of start is sent to external applications, which is then delayed.
 1. The check can time out if the network is very busy, which then stops the simulation even though the simulation might be proceeding perfectly fine.

I will close this PR, and replace with one tied to other issues in the synapse expander.

PR #350, opened 2018-10-03
Fix chip power monitor issues

andrewgait:
Interface to chip power monitor was broken, have fixed it.


alan-stokes:
i know this has already gone in, but why is chip power monitor a system binary? it takes simulation_utilities for timings. it should be responding to simulation control logic. The c code has the resume functionality, the auto_pause and resume and everything it needs. Putting it in system field will most likely break it. for one, itll be stuck in sync i think

andrewgait:
The change moved it from being a system binary to a simulation... having it as system was what was breaking it.

alan-stokes:
ah. must have misread it due to tea downing. no worries. im refueling now. sorry for confusion. im amazed it ever was system....... must have been a brain fart back then as well

PR #355, opened 2018-10-10
Citation

jspaaks:
Hi @alan-stokes, 
I tried validating against the CFF schema 1.0.3 and found a couple of errors, corrected with this PR. BTW I used ``cffconvert`` from PyPI to do the validation, using:
```bash
cd <the spinnaker dir>
cffconvert --validate
# (usually) results in bunch of nearly illegible text being emitted by
# pykwalifire. Only if it's completely quiet, your file is valid.
```
I edited the CFF until it passed validation (https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/commit/551e957d30fe30c886ac3bbb5a292d193ae8f5f6), then reintroduced some errors which are due to errors/incompleteness in the CFF schema, not in your file. Those errors have been fixed in bleeding edge, but not yet in stable. For reference:
- https://github.com/citation-file-format/schema/pull/2
- https://github.com/citation-file-format/schema/pull/4

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,68 +2,68 @@
 authors: 
   - 
     affiliation: "University Of Manchester"
-    email: alan.stokes-2@manchester.ac.uk
+#    email: alan.stokes-2@manchester.ac.uk
     family-names: Stokes
     given-names: Alan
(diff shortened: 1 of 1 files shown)


alan-stokes:
thanks rowley

PR #356, opened 2018-10-12
Fix command sender machine vertex

rowleya:
Fixes command sending to work with both application and machine vertices


alan-stokes:
looks good. waiting on ticks

PR #357, opened 2018-10-15
Fix finisher

rowleya:
Fixes issue where the incorrect state is looked for on application exit


alan-stokes:
im still inclined to bounce this until we have a better solution to the whole simulation state machine issue.

PR #361, opened 2018-10-23
Allow buffer manager to use sqllite

Christian-B:
off by default but now settable using config.


Christian-B:
Most changes here changed again in  https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/363

PR #363, opened 2018-10-24
Buffer cleanup

Christian-B:
OK to merge.

- Includes but overwrites much of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/361 (allow user to use sql Lite) so closed that one

- get data is now by placement (which has a vertex) rather than by vertex where the placement must then be found.

- get data directly from BufferedReceivingData rather than getting a pointer and then getting the data.
(avoids the need for a database pointer Object)

Must be done at the same time as: https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/95 and https://github.com/SpiNNakerManchester/sPyNNaker/pull/565
....

### DIFF
--- a/fec_integration_tests/test_buffer_manager_listener_creation.py
+++ b/fec_integration_tests/test_buffer_manager_listener_creation.py
@@ -1,3 +1,5 @@
+import os
+import tempfile
 import unittest
 from spinn_utilities.overrides import overrides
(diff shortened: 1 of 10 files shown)


Christian-B:
Now part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/370
So closing separate PR.

PR #369, opened 2018-11-10
Fix run forever

rowleya:
Fairly minor fixes that just make things a bit clearer.  This also avoids extracting buffer data when there can't be any!

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -2016,8 +2016,7 @@ class AbstractSpinnakerBase(SimulatorInterface):
         if not self._use_virtual_board and n_machine_time_steps is not None:
             algorithms.append("BufferExtractor")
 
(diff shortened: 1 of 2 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2010,7 +2013,7 @@ def _create_execute_workflow(
 
         # ensure we exploit the parallel of data extraction by running it at\
         # end regardless of multirun, but only run if using a real machine
-        if not self._use_virtual_board:
+        if not self._use_virtual_board and n_machine_time_steps is not None:
```
alan-stokes: the reasoning for this, is that n_machine_time_steps is only set to None if the run_time is none, and its a explicit param setting for run forever. whereas n_machine_time_steps is a indirect marker. I had to traverse the code to ensure that the only way n_machine_time_steps was set to none was in run_forever

```suggestion
        if not self._use_virtual_board and run_time is not None:
```

PR #375, opened 2018-12-04
Map database

Christian-B:
Builds on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/370

Works with Java main but ignores extra monitors


alan-stokes:
failed tests

Christian-B:
Alan Travis tells us when tests fail! We don't need another email from you!

PR #376, opened 2018-12-05
Ds database

Christian-B:
REVIEW DO NOT MERGE!


Christian-B:
requires https://github.com/SpiNNakerManchester/DataSpecification/pull/52
causes
https://github.com/SpiNNakerManchester/sPyNNaker/pull/572 to be needed.

Christian-B:
Oops note done DSGRegionReloader

That will need looking at!

Christian-B:
SynapseExpander is also a seperate process which does not share data.

Discussions with @alan-stokes and @rowleya suggest this may never be converted to Java

alan-stokes:
tis coz there be no use. it cannot be parallised more than it already is

alan-stokes:
looks good to me

PR #377, opened 2018-12-06
Update equivalent vertex check

andrewgait:
Fix InsertExtraMonitorVerticesToGraphs interface function to run in seconds rather than minutes or hours (was seemingly at least quadratic with number of boards used in job).

[The branch name is slightly misleading, apologies for this as it was named for the first quick fix I tried, not the actual fix that's in here now...]

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
+++ b/spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
@@ -71,37 +71,37 @@ class InsertExtraMonitorVerticesToGraphs(object):
 
         for chip in progress.over(machine.chips):
             if not chip.virtual:
-#                 machine_vertex = self._exists_equiv_vertex(
-#                     chip.x, chip.y, machine_graph,
-#                     ExtraMonitorSupportMachineVertex)
-#                 if machine_vertex is None:
-                # add to machine graph
-                machine_vertex = self.__new_mach_monitor(chip)
-                machine_graph.add_vertex(machine_vertex)
+                machine_vertex = self._exists_equiv_vertex(
+                    chip.x, chip.y, vertex_to_chip_map,
+                    ExtraMonitorSupportMachineVertex)
+                if machine_vertex is None:
+                    # add to machine graph
+                    machine_vertex = self.__new_mach_monitor(chip)
+                    machine_graph.add_vertex(machine_vertex)
 
-                vertex_to_chip_map[chip.x, chip.y] = machine_vertex
-                extra_monitor_vertices.append(machine_vertex)
+                    vertex_to_chip_map[chip.x, chip.y] = machine_vertex
+                    extra_monitor_vertices.append(machine_vertex)
 
-                # add application graph as needed
-                if application_graph is not None:
-#                     equiv_vertex = self._exists_equiv_vertex(
-#                         chip.x, chip.y, application_graph, ExtraMonitorSupport)
-#                     if equiv_vertex is None:
-                    app_vertex = self.__new_app_monitor(chip)
-                    application_graph.add_vertex(app_vertex)
-                    graph_mapper.add_vertex_mapping(
-                        machine_vertex, Slice(0, 0), app_vertex)
+                    # add application graph as needed
+                    if application_graph is not None:
+                        equiv_vertex = self._exists_equiv_vertex(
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -72,35 +72,35 @@ def _handle_second_monitor_functionality(
...
 
     @staticmethod
-    def _exists_equiv_vertex(x, y, graph, vertex_type):
-        for vertex in graph.vertices:
-            if isinstance(vertex, vertex_type) and any(
-                    constraint.x == x and constraint.y == y
-                    for constraint in locate_constraints_of_type(
-                        vertex.constraints, ChipAndCoreConstraint)):
-                return vertex
+    def _exists_equiv_vertex(x, y, map, vertex_type):
```
alan-stokes: map is a shadow name. needs changing

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -72,35 +72,35 @@ def _handle_second_monitor_functionality(
...
+        if (x, y) in map and isinstance(map[x, y], vertex_type) and any(
```
alan-stokes: if your using the vertex_to_chip_map, you can ensure all verts in it are of extra monitor type. so the check is redundant.

alan-stokes: this becomes 
```suggestion
        if (x, y) in map:
            return map[x,y]
        return None
```

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -129,7 +129,8 @@ def _handle_data_extraction_vertices(
             # add to application graph if possible
             if application_graph is not None:
                 equiv_vertex = self._exists_equiv_vertex(
-                    ethernet_chip.x, ethernet_chip.y, application_graph,
+                    ethernet_chip.x, ethernet_chip.y,
+                    vertex_to_ethernet_connected_chip_mapping,
                     DataSpeedUpPacketGather)
```
alan-stokes: remove as you dont need the type now
```suggestion
                    )
```

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -72,35 +72,35 @@ def _handle_second_monitor_functionality(
...
-                        graph_mapper.add_vertex_mapping(
-                            machine_vertex, Slice(0, 0), app_vertex)
+                    vertex_to_chip_map[chip.x, chip.y] = machine_vertex
+                    extra_monitor_vertices.append(machine_vertex)
+
+                    # add application graph as needed
+                    if application_graph is not None:
+                        equiv_vertex = self._exists_equiv_vertex(
+                            chip.x, chip.y, vertex_to_chip_map,
+                            ExtraMonitorSupport)
```
alan-stokes: remove as you dont need the type
```suggestion
                            )
```

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -144,7 +145,8 @@ def _handle_data_extraction_vertices(
                     machine_vertex = equiv_vertex.machine_vertex
             else:
                 machine_vertex = self._exists_equiv_vertex(
-                    ethernet_chip.x, ethernet_chip.y, machine_graph,
+                    ethernet_chip.x, ethernet_chip.y,
+                    vertex_to_ethernet_connected_chip_mapping,
                     DataSpeedUpPacketGather)
```
alan-stokes: remove as not need type
```suggestion
                    )
```

alan-stokes:
having investigated and discussed with @Christian-B the fact that the app graph is nuked in states where recycling the graph is needed. 

so all this safety code can go bye bye due to safety code in ASB in relation to the matching between self._orginial_app_graph vs self._app_graph. 

least i had the foresight to add that safety code in beforehand. lol.  just not the foresight to realise this slow safety is thus redundant. thank you for making me go down this road

PR #378, opened 2018-12-11
fix command sender

rowleya:
Removes delays between packets from the command sender.  It is assumed that this will already have appropriate delays between packets (commands can have delays between sending anyway).

### DIFF
--- a/c_common/models/command_sender_multicast_source/src/command_sender_multicast_source.c
+++ b/c_common/models/command_sender_multicast_source/src/command_sender_multicast_source.c
@@ -29,8 +29,6 @@ static uint32_t n_timed_commands;
 static uint32_t n_start_resume_commands;
 static uint32_t n_pause_stop_commands;
 static uint32_t next_timed_command;
-static uint32_t expected_next_send;
-static uint32_t time_between_sends;
 static bool resume = true;
 
 //! values for the priority for each callback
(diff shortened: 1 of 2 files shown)


alan-stokes:
thought the delay was to try to avoid a overload on routing. not to hit a peak. by delaying a chunk, youd be more likely to miss peaks in transmission. by removing, how are you going to ensure this?

rowleya:
This is a command sender - unless you flood a specific time step with commands, you won't need to delay things (and then it probably won't work anyway).  Additionally, if you send some commands repeatedly, there is already a delay between time steps.  Bearing in mind, the most common use case of this vertex is to send a start command at the start and an end command at the end, it is unlikely to overload the router.

Note that as it stands in master, it is not possible for this class to finish.  This is because the timer has been stopped, but it is still using the timer to delay the packets.

alan-stokes:
think ya missing what im hinting at. flooding the network from other components. aka like a SSA firing 256 packets at the same timestep, repeat over the machine. That generates a peak of traffic which could drown out and make drop these commands. thus why delaying them a bunch during retires may be usful. 

unknown about class unable to finish. that seems like a bug. but then im not digging into it deeply at the moment.

alan-stokes:
delay within the time step, as doing one per timestep makes the external device start possibly delayed beyond the starting millisecond. which coiuld have adverse effects on plastic synapses based off the retina packets etc

andrewgait:
That's true during the run, but I think this is getting at the problem after the run is "over": the timer has stopped, so if it keeps on being fed packets to delay further, then it's never going to finish because the timer can't increment any more to catch up... ?

[Possibly related: Shouldn't an "end" command override absolutely everything else?  Otherwise you're asking for trouble...]

rowleya:
It is possible to send messages during the run, but most of those will already be spread out, and we have no control over those in any case.  The command sender sends very little in all the use cases we have seen.  It could be made to spread out those packets but the effort involved in making it work like that only during simulation is quite a lot, and could be done if this causes issues in the future.  I think we only added it in here because it was being done elsewhere...

PR #380, opened 2018-12-20
random tx delay (0-0.5 x time step)

rjames91:
Added a random timer tick callback delay for spike source arrays that differs across cores. The delay factor is generated on host (0-0.5) and is then multiplied on each core by the number of microseconds in a timer tick.

The accum/uint union is a hacky way of getting the right number read back on the core - I'm sure there's a more sensible way of doing this...


rowleya:
I have updated this to the code that we have used elsewhere to do a similar thing.  Let me know if this still works, or if it now breaks things...

PR #394, opened 2019-02-19
SQLite doesn't really have INT

dkfellows:
Of the two, it turns out that `INTEGER` is the one that the SQL standard actually specifies.


dkfellows:
(where “standard” means SQL-92)

PR #395, opened 2019-02-27
No more layout enums

dkfellows:
C has good ways to represent data structures in memory so that you virtually never need to do complex address arithmetic (especially when everything is word-aligned). This commit is about converting our code to actually use those techniques.

The only one of our protocols that is resistant to this approach is EIEIO, because that was written by someone truly twisted.


dkfellows:
Note that this PR must not be merged without _thorough_ testing.

rowleya:
Having merged in master, this is causing RTEs and other failures.  See:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/sPyNNaker8%20Integration%20Tests/detail/no-more-layout-enums/5/pipeline/

dkfellows:
#462 effectively supersedes this.

Issue #396, opened 2019-02-28
write_reload_steps


Christian-B:
The only use of this config value is:
abstract_spynakker_base
        if self._config.getboolean("Reports", "write_reload_steps"):
            logger.warning("Reload script is not supported in this version")

NUKE!

alan-stokes:
merged. so closing

PR #397, opened 2019-02-28
NUKED write_reload_steps config value!

Christian-B:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/396

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -596,10 +596,9 @@ class AbstractSpinnakerBase(SimulatorInterface):
 
     # options names are all lower without _ inside config
     DEBUG_ENABLE_OPTS = frozenset([
-        "reportsenabled", "displayalgorithmtimings",
+        "reportsenabled",
         "clear_iobuf_during_run", "extract_iobuf", "extract_iobuf_during_run"])
     REPORT_DISABLE_OPTS = frozenset([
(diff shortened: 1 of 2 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -596,10 +596,9 @@ def _load_config(self, filename, defaults, validation_cfg):
 
     # options names are all lower without _ inside config
     DEBUG_ENABLE_OPTS = frozenset([
-        "reportsenabled", "displayalgorithmtimings",
```
dkfellows: What's going on here?

Christian-B: When the user sets mode = debug ASB (or now ConfigHandler) looks for config settings to change from false to true.

Similarly if reportsEnabled=False it looks for ones to set false.

The way this is done is to either
1. Look if the config name has "write" in it
2. The config name is in this list.

On a side note DEBUG_ENABLE_OPTS should be moved to ConfigHandler

dkfellows: So (the equivalent of) `displayalgorithmtimings` has moved to config handler? That's fine.

Christian-B: No in this case the config value "displayalgorithmtimings" was no longer used anywhere.

So what was the reason for setting it to true other than confusion!

PR #408, opened 2019-03-22
Sane DSR API types

dkfellows:
The types in the data specification access API “leave something to be desired”. In particular, `data_specification_get_data_address()` returns a supposedly raw address (despite it actually being a pointer to a very specifically laid out piece of memory) and `data_specification_get_region()` returns a value that probably ought to be cast to something other than a raw address straight away (as it points to whatever the Python code said that it should), and so that ought to be a `void *`.

This makes this true (and makes `data_specification_get_region()` be `static inline` as it is trivial in terms of instructions issued).

----

Because the code doesn't use “raw addresses” everywhere so much any more, I've changed some of the variable names to match.


rowleya:
Is this still relevant after data speed up merges?

PR #409, opened 2019-03-29
C version of router compressor with bit field

alan-stokes:
this holds most of the functional work for merging bit-fields into the routers. It should be general purpose. Based on APIs. The only place that generates these bit-fields is currently spynnaker though. 

currently in development


c_common/models/Makefile
```diff
@@ -1,6 +1,8 @@
-BUILD_DIRS = command_sender_multicast_source reverse_iptag_multicast_source \
-	live_packet_gather chip_power_monitor data_speed_up_packet_gatherer \
-	extra_monitor_support
+BUILD_DIRS = compressor_bit_field_models \
```
dkfellows: Why are the two `*_multicast_source` models not on the build list any more?

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
```diff
@@ -534,6 +536,11 @@ def _write_configuration(self, spec, n_machine_time_steps):
         # write SDP port to which SDP packets will be received
         spec.write_value(data=self._receive_sdp_port)
 
+        # write random tx offset in microseconds
+        import numpy as np
+        tx_offset = 0.5 * np.random.rand()
```
dkfellows: For reproducibility, the random source needs to be seedable and independent of other uses of random numbers.

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
```diff
@@ -79,7 +81,7 @@ class ReverseIPTagMulticastSourceMachineVertex(
     #          5. has key, 6. key, 7. mask, 8. buffer space,
     #          9. send buffer flag before notify, 10. tag,
     #          11. tag destination (y, x), 12. receive SDP port)
-    _CONFIGURATION_REGION_SIZE = 12 * 4
+    _CONFIGURATION_REGION_SIZE = 13*4  # 12 * 4
```
dkfellows: Please update the comment just above this line too.

c_common/models/Makefile
alan-stokes: have chekced and all algorithms have them. so they must have disappeared and then got readded. given this pr is dead. assuming nothing to worry about

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
alan-stokes: also fixed in all algiorithms

alan-stokes:
killing as out of date and protentially buggy whereas all algorithms has all fixes currently known about

PR #421, opened 2019-05-10
added _ne__ for the dinasours still using python 2.7

Christian-B:


alan-stokes:
postponed till discussion about 
https://stackoverflow.com/questions/4352244/python-should-i-implement-ne-operator-based-on-eq

is determined to a awnser

PR #432, opened 2019-05-24
Make the recording implementation produce no warnings when compiled

dkfellows:
Also makes names clearer. And fixes a masking bug.

And use `static` because that really does stop names from being public outside the “compilation unit” (the file).

With this, we can start to contemplate adding `-Werror` to the CFLAGS our makefiles, which makes every warning fatal.


dkfellows:
Not enough added value as its own thing

PR #434, opened 2019-05-28
Actually make sure an individual word gets sent during speed up protocol

andrewgait:
Fix for https://github.com/SpiNNakerManchester/sPyNNaker/issues/652; in this particular case, the bug was only present with advanced monitor support turned on; in the particular example, the final message sent using the data speed up protocol was only one word in size, and was not being sent by the monitor core to the speed up core - and so the final data value was not being filled in correctly.


alan-stokes:
looks fine to me. surprised it only shows up in that specific situation

andrewgait:
I'm surprised too - I tried to replicate it, but it's dependent upon a mixture of runtime, n_neurons and how many sequence numbers end up being used...

PR #436, opened 2019-05-31
All algorithms

alan-stokes:
tie in for the project to test mapping algorithms. is a blind copy of c model with bitfields


alan-stokes:
only errors are now on c code. which @Christian-B is fiddling with. so the token for fixing things here is handed over to @Christian-B. But I may steal it back if the spynnaker8 integration tests (https://github.com/SpiNNakerManchester/sPyNNaker8/pull/240) fail on something i can fix quickly.

c_common/models/compressor_bit_field_models/src/compressor_includes/remove_default_routes.h
```diff
@@ -0,0 +1,114 @@
...
+    // Mark the entries to be removed from the table
+    bit_set_t remove;
+    bool success = bit_set_init(&remove, routing_table_get_n_entries());
+    if (!success) {
+        log_debug("failed to initialise the bit_set. shutting down");
+        return false;
+    }
+
+    // Work up the table from the bottom, marking entries to remove
+    for (int i = routing_table_get_n_entries() - 1; i < 0; i--) {
```
dkfellows, line 41: This line looks suspicious to me. I strongly doubt that `routing_table_get_n_entries()` will be returning a negative number!

alan-stokes: thats interesting. given thats not in the code we worked on. thats mundy's code..... but your right. it looks like the bool operator is reversed for some reason from what we'd expect......

might just be a use of switching the bool operator and seeing if it breaks anything

alan-stokes: given the orignal code was even more odd

https://github.com/project-rig/rig_routing_tables/blob/master/include/remove_default_routes.h#L15

im just testing with the correct bool logic to see if it breaks anything on the local machine

Christian-B: Alan’s attempt at converting Dogie Mindy code that counted on uint32_t overflow to work.

Luckily not in a critical Bit of code.

alan-stokes: fixed

dkfellows:
I've never cared for the name “all algorithms”, but the idea of filtering seems OK.

My remaining quibbles with the new memory allocator code absolutely will not block this being merged!

alan-stokes:
the name "all algoritrhms" came from when this was being built to try to fix roberts issue. where i kept adding and adding stuff and it just became a accum of stuff like "bit-field_expander, spreader_placer, pair key alloc, bit_field_compressor". i admit what its doing now is alas not too descriptive. but too much faff to rename it all

PR #438, opened 2019-06-10
Part rig

Christian-B:
Part of https://github.com/SpiNNakerManchester/PACMAN/pull/221


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1474,10 +1474,10 @@ def _do_mapping(self, run_time, n_machine_time_steps, total_run_time):
...
                     "Reports", "write_router_summary_report"):
                 algorithms.append("RouterSummaryReport")
-            if self._config.getboolean(
-                    "Reports", "write_compressed_router_summary_report") and \
-                    self.use_virtual_board:
-                algorithms.append("CompressedRouterSummaryReport")
+            # if self._config.getboolean(
+            #        "Reports", "write_compressed_router_summary_report") and \
+            #        self.use_virtual_board:
+            #    algorithms.append("CompressedRouterSummaryReport")
```
rowleya: Does this cause issues?  I was added in a previous branch but is disabled by default...

Christian-B: it can out of memory in a big network. But the correct fix is to remove it from the cfg not asb

PR #440, opened 2019-06-11
Python 3: Revenge of the divide monster

andrewgait:
Trying to use the profiler with machine_time_step=100 (i.e. timestep=0.1) led to some crazy memory errors.  I'm not sure looking at the history of this code how this ever actually worked in the first place.


andrewgait:
I think it will be better done with numpy.linspace, yes - I'll look into it.

I'm pretty sure people have used the profiler (on neurons) with the micro-cortical column at machine_time_step_ms=0.1 (and other values < 1.0), which is why I'm surprised this hasn't been seen before.

PR #442, opened 2019-06-14
Sanity for router timeouts!

dkfellows:
* Avoid infinite timeouts, which might be buggy.
* Pass the timeouts as tuples (individual values don't make sense).
* Give names to the timeouts used.


alan-stokes:
really worried when youve put a long timeout which isnt infinite. the whole logic of the protocol is on the premise that the blocking nature of the infinite timeouts. by making it potentially lossy, your making the protocol flaky

PR #451, opened 2019-07-03
Fix sizes of messages written and add a timeout to the gatherer

dkfellows:
Timeouts are good. Timeouts mean that even if a bunch of messages get lost _including the last message in a stream_, we'll eventually recover.

Writing the correct size of message is even better! Using the right message sizes (instead of their negation) means that we don't crash SCAMP when we send response messages asking for some packets to be resent.

### DIFF
--- a/c_common/models/data_speed_up_packet_gatherer/src/data_speed_up_packet_gatherer.c
+++ b/c_common/models/data_speed_up_packet_gatherer/src/data_speed_up_packet_gatherer.c
@@ -180,7 +180,8 @@ typedef struct data_out_config_t {
 typedef enum callback_priorities {
     MC_PACKET = -1,
     SDP = 0,
-    DMA = 0
+    DMA = 0,
+    TIMER = 1
 } callback_priorities;
 
 // Note that these addresses are *board-local* chip addresses.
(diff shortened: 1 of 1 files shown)


c_common/models/data_speed_up_packet_gatherer/src/data_speed_up_packet_gatherer.c
```diff
@@ -359,20 +359,42 @@ static inline void set_message_length(const void *end) {
...
 static void process_missing_seq_nums_and_request_retransmission(void) {
+    //! \brief Used to guard access to the received_seq_nums_store from this
+    //!   function; it counts the number of running calls to this function.
+    //!   Access to this variable is only allowed when you have disabled
+    //!   interrupts!
+    static uint access_lock = 0;
+
+    uint sr;
+    sr = spin1_irq_disable();
+    if (++access_lock == 0) {
```
rowleya: Is this correct?  I am looking at this and thinking that if adding 1 to an unsigned integer == 0, then value must have been 0xFFFFFFFF previously...

c_common/models/data_speed_up_packet_gatherer/src/data_speed_up_packet_gatherer.c
```diff
@@ -324,51 +340,93 @@ static inline uint calculate_sdram_address_from_seq_num(uint seq_num) {
...
 static void process_missing_seq_nums_and_request_retransmission(void) {
+    //! \brief Used to guard access to the received_seq_nums_store from this
+    //!   function; it counts the number of running calls to this function.
+    //!   Access to this variable is only allowed when you have disabled
+    //!   interrupts!
+    static uint access_lock = 0;
+
+    uint sr;
+    sr = spin1_irq_disable();
+    if (++access_lock == 1) {
```
rowleya: OK, I think this now means you can never get in to the function.  Should this be a post-increment, so you are looking at the value before it was incremented in the comparison?

Issue #452, opened 2019-07-08
Better tag reporting


dkfellows:
It would be very nice if there was a report that read the configuration of the tags off the machine. Yes, we know what we asked the machine to set, but we can't always safely assume that that's exactly what was actually done. A new report would help a lot, especially with debugging.

If the report is run after the run, it would be great if it returned the number of packets sent via the tag. Apparently this is a capability of ybug (but stripped in SpiNNMan), and it would be a great thing to describe in a report. It might also be a sane thing to report as provenance.

dkfellows:
See also https://github.com/SpiNNakerManchester/SpiNNMan/issues/169

Christian-B:
there is a report tags_on_machine.tx

PR #454, opened 2019-07-09
Simple reporting of actual tags on machine

dkfellows:
This was my attempt to implement #452, but I don't think it is complete or correct. (It was enough to let me debug the final problem in Data In, but the reporting the rest of the time is imperfect.) Please feel free to improve it!


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1036,6 +1037,8 @@ def _run_algorithms(
             self._txrx = executor.get_item("MemoryTransceiver")
             self._machine_allocation_controller = executor.get_item(
                 "MachineAllocationController")
+            report_folder = executor.get_item("ReportFolder")
```
Christian-B: Why not just use
self._report_default_directory

That is setup in the __init__

dkfellows: Feel free to change the code!

PR #455, opened 2019-07-11
bits needed for sdram edges

alan-stokes:
https://github.com/SpiNNakerManchester/PACMAN/pull/232


alan-stokes:
so this is dead. i think the changes are the same. but they been cherry picked from roberts changes in 
https://github.com/orgs/SpiNNakerManchester/projects/22

so this pr and project getting killed for clarity

PR #457, opened 2019-07-15
Unordered c

Christian-B:
minimiser in c


Christian-B:
replaced with to keep branches simpler.

PR #459, opened 2019-07-19
Ear bit fields cleaned

alan-stokes:
stuff to allow roberts ear to work correctly within pynn without hacks.

also adds support for static projections and will include selective recording. needs to be considered carefully for plastic projections.

needs prs:


dkfellows:
We'll not have this finished by the 6.0 release

Christian-B:
While this is nice work it is judged as too big an API change for the benefit.

Has also now fallen behind much of the other PRs based off the work here.

Closing PR but keeping branch for historical and future reference reasons.

PR #462, opened 2019-07-24
Function scope

dkfellows:
Limit the scopes of functions and _partially_ struct-ify things.


dkfellows:
This is the general cleanup I think is warranted for the C code.

PR #478, opened 2019-08-29
Ensure that report directories are unique

dkfellows:
If report directories aren't unique (which can happen due to annoyingness with clock granularity in parallel runs), there's a potential for catastrophe when databases get shared (which happens annoyingly often during integration testing).

Don't do that. Really.


Christian-B:
Rubbered and not in the sense of rubber mallet!

PR #486, opened 2019-09-19
Constant byte size multipliers

dkfellows:
This branch defines
```python
BYTES_PER_WORD = 4
BYTES_PER_SHORT = 2
BYTES_PER_KB = 1024
```
and updates the rest of the code (downstream from here) to use them so that it is clearer what is actually in bytes and what just happens to be multiplied by these values.


alan-stokes:
we may need to be careful with this one. As ive done the same thing during bitfields, roberts changes etc. so there's 

# conversion from words to bytes
WORD_TO_BYTE_MULTIPLIER = 4
# converts between micro and milli seconds
MICRO_TO_MILLISECOND_CONVERSION = 1000

BITS_PER_WORD = 32.0

that ive added. these may clash with these changes a wee bit

dkfellows:
Having bits per word be a floating point number is just a little bit odd…

PR #491, opened 2019-09-25
cherry picked for neuron recorder

alan-stokes:
changes from roberts conversion for recording floats, doubles, int's and encapsulating all the selective recording stuff. 

includes some magic number removals as well, and mini bug fixes for iobuf extractor. 

depends upon spynnaker branch
https://github.com/SpiNNakerManchester/sPyNNaker/pull/706****
https://github.com/SpiNNakerManchester/DataSpecification/pull/71
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/126

spynnaker8 test pr
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/268


c_common/front_end_common_lib/src/recording.c
```diff
@@ -59,7 +59,7 @@ struct recording_data_t {
     uint32_t time_between_triggers;
     uint32_t last_sequence_number;
     recording_channel_t *region_pointers[0];
-};
+} recording_data_t;
```
andrewgait: For consistency this should be the same as the struct above it i.e. there's a missing "typedef" here.

c_common/front_end_common_lib/src/recording.c
```diff
@@ -538,7 +544,8 @@ static void recording_dma_finished(uint unused, uint tag) {
 }
 
 bool recording_initialize(
-        address_t recording_data_address, uint32_t *recording_flags) {
+        address_t recording_data_address, uint32_t *recording_flags,
+        uint32_t *words_read) {
     struct recording_data_t *recording_data =
```
andrewgait: Not really a comment on this change per se - I think it's fine - but there's some calls of this function elsewhere (e.g. spinnaker_graph_front_end/examples/hello_world/src/hello_world.c line 96, and probably other GFE examples too) that need to be edited to match this change.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1998,8 +2007,9 @@ def _recover_from_error(self, exception, exc_info, executable_targets):
                 failed_cores = self._txrx.get_cores_not_in_state(
                     self._executable_types[executable_type],
                     executable_type.end_state)
-                for (x, y, p), core_info in failed_cores:
-                    unsuccessful_cores.add_processor(x, y, p, core_info)
+                for (x, y, p) in failed_cores:
+                    unsuccessful_cores.add_processor(
+                        x, y, p, failed_cores.get_cpu_info(x, y, p))
```
andrewgait: Won't this be slower than what it's replacing, given that get_cores_not_in_state(...) does this work for you already?

spinn_front_end_common/utilities/constants.py
```diff
@@ -21,12 +21,14 @@
 SPIKE_INJECTOR_CORE_APPLICATION_ID = 0xAC9
 
 BITS_PER_WORD = 32.0
+BYTES_PER_WORS = 4
```
andrewgait: Should be "BYTES_PER_WORDS" (and similarly on sPyNNaker branch).  Also, either both of these constants should be floats or both of these constants should be integers; probably the latter?

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -303,6 +304,19 @@ def convert_vertices_to_core_subset(vertices, placements):
     return core_subsets
 
 
+def find_executable_start_type(machine_vertex, graph_mapper=None):
```
andrewgait: I can see how this function would be useful, but is it actually called from anywhere at the moment?

c_common/front_end_common_lib/src/recording.c
alan-stokes: buger. just noticed i dont even have  GFE repo in this set of the tools...... bugger. will solve once the time stuff is done

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: the issue here is some change (which might have been fixed in master, but this worked) where the CoreInfo or CoreSubsets (which ever was replaced) doesnt have the right api and data to give us what we need for the failure state (mainly iobuf extractor)

spinn_front_end_common/utilities/constants.py
alan-stokes: if theres multipl;e of these. then thats a failure on its own. (there should only be a version of this in FEC. no point having the same constants elsewhere.

on floats vs ints. there was a reason to make the bits a float. i think its to do with a numpy expectation. id need to change it and review. will do so after timing

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: ermmmm...... having checked from the ear bitfield's cleaned branches. it seems that's a hang over from the all algorithms. its used in "find application chips used" and "bit_field_compressor_report" so i think its safe to remove it from this version. But its worth noting for its going to cause a merge confict when these divergant bits say hi to each other

alan-stokes: solved this one quickly

spinn_front_end_common/utilities/constants.py
alan-stokes: ok having investigated. 1. theres 2 params for the same thing. BYTES_PER_WORDS and Bytes_To_WORD_MULTIPLIER, so reduced them. But the issue with flaots vs ints is as follows:
1. the BitsToWord is used to determine bitfield size mostly. and that maths needs a math.ceil of a float, so this removes the casting from int to float all the time. 
2. the BYtes_per_word is used often to find size of a memory region. which requires a int, so by keeping it a int, you avoid more casts.

so summary theyre different types to avoid castings more often for each use case

alan-stokes:
to reply to your question @andrewgait 
the SSa and SSP implement their own recordings down in the c and python levels. Niehter of them support indexing (the python will raise an exception currently [and is the same in master, so haven't removed functionality here] saying that index'ing is not support in the SSA and SSP.) 

and so these functionalities were never made to access the neuron recorder either. I think the problem is aligned to handling multiple spikes per time step. Im inclined to think that it should be possible to use the API, but Id feel that's another work for another pr. 

the weird test failing we think is SCAMP related, so no.

c_common/front_end_common_lib/src/recording.c
alan-stokes: wonder how i missed that. fixed in last commit. wonder why the compiler was happy. lol

c_common/front_end_common_lib/src/recording.c
alan-stokes: fixed. new pr and added to all pr links

spinn_front_end_common/utilities/constants.py
andrewgait: OK, that's fine - thanks.

c_common/front_end_common_lib/src/recording.c
andrewgait: I don't think the compiler necessarily cares in this case, but it seemed sensible to suggest doing the same thing for each struct.

andrewgait:
(Assuming you fix the flake8 issue that's just turned up, of course :-) ).

alan-stokes:
flake 8..........what flake 8? i see nothing ;-)

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -440,8 +442,8 @@ def __init__(
         self._mapping_tokens = None
         self._load_outputs = None
         self._load_tokens = None
-        self._last_run_outputs = None
-        self._last_run_tokens = None
+        self._last_run_outputs = list()
+        self._last_run_tokens = list()
```
rowleya: Why do these need to be empty lists?

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1699,6 +1703,7 @@ def _do_load(self, graph_changed):
             self._config.getboolean("Reports", "write_memory_map_report") and
             graph_changed
         )
+        inputs["NoSyncChanges"] = self._no_sync_changes
```
rowleya: It isn't clear why this is needed

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1771,6 +1776,7 @@ def _do_load(self, graph_changed):
         executor = self._run_algorithms(
             inputs, algorithms, [], tokens, required_tokens, "loading",
             optional_algorithms)
+        self._no_sync_changes = executor.get_item("NoSyncChanges")
```
rowleya: It isn't clear why this is needed

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
```diff
@@ -673,16 +704,6 @@ def is_recording(self):
...
-        """ Updates the buffers on specification of the first machine timestep.
-            Note: This is called by injection.
-
-        :param first_machine_time_step:\
-            The first machine time step in the simulation
-        :type first_machine_time_step: int
-        :param run_until_timesteps:\
-            The last machine time step in the simulation
-        :type run_until_timesteps: int
-        """
```
rowleya: I think this was removed by the merge for some reason...

spinn_front_end_common/utilities/globals_variables.py
```diff
@@ -90,3 +94,20 @@ def set_failed_state(new_failed_state):
...
+        if _cached_simulator is not None:
+            return _cached_simulator.get_generated_output(output)
+        else:
+            raise ValueError(
+                "You need to have ran a simulator before asking for its "
+                "generated output, and the simulator needs to be cached "
+                "before you can request outputs.")
+    else:
+        raise ValueError(
+            "There should be some sort of simulator set. Why am i here?!")
```
rowleya: It isn't clear who uses this

c_common/front_end_common_lib/src/recording.c
```diff
@@ -633,6 +642,11 @@ bool recording_initialize(
     simulation_dma_transfer_done_callback_on(
             RECORDING_DMA_COMPLETE_TAG_ID, recording_dma_finished);
 
+    // update word tracker for anything else
+    *words_read += (
+        (sizeof(recording_data_t) / WORD_TO_BYTE_CONVERSION) +
+        (N_WORDS_USED_OUTSIDE_STRUCT_PER_REGION * n_recording_regions));
+
```
rowleya: This might be better as bytes read, then you don't need to do so much work here.  Another possibly simpler option is to pass in a ```void **``` or a ```uint8_t **``` as the address to read from, then you can just update the pointer to the region.

-

c_common/front_end_common_lib/src/recording.c
```diff
@@ -92,6 +92,12 @@ static read_request_packet_data *data_ptr;
 //! The time between buffer read messages
 #define MIN_TIME_BETWEEN_TRIGGERS 50
 
+//! n words outside struct per recording region used
+#define N_WORDS_USED_OUTSIDE_STRUCT_PER_REGION 2
```
rowleya: What are these mysterious words?

alan-stokes: these are actually the size of the recording_channel_t. I admit it should be a sizeof instead of a magic constant. Guess i didnt realise what it was, wehn we did this.

why they are at the bottom of the recording_data_t is beyond me, but i think tis to do with the buffer manager needing to know where to read.

c_common/front_end_common_lib/src/recording.c
alan-stokes: reason for words instead of bytes, is that the guy using it is using it as an array offset. someone needs to do the divide. up to you whom. 

I guess you could do a return (param out) for the address and then the neuron recorder would be able to use it without the offset..... youd need to change everywhere who reads recording_init to change the param type again..... but doable.

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
alan-stokes: im passing this one

spinn_front_end_common/utilities/globals_variables.py
alan-stokes: if im honest. i dont know. but my only think is this might be a feature creep for making projection data extraction safer. if you asked for it before you ran. But i honestly dont know. one to find out. pull out and rerun tests.

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: this is a side effect of the get_generated_output. so that you dont need to check for None before asking for in.

spinn_front_end_common/utilities/globals_variables.py
alan-stokes: this whole stuff feels like at some point we were asking for something before it was filled in, and so i was adding safety code to ensure the error was readable. but i dont know if its something still applicable

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: i think this is a slip from having bitfield stuff inside the ear branches. verification might be to chekc if this is in all_algorithms.

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: i think this is a slip from having bitfield stuff inside the ear branches. verification might be to chekc if this is in all_algorithms.

spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
rowleya: Looks like a merge issue, resolved.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Yes, looks like it - removing!

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Removed

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: This also comes from another branch, so removing from here for now...

spinn_front_end_common/utilities/globals_variables.py
rowleya: Removed as from elsewhere

c_common/front_end_common_lib/include/recording.h
```diff
@@ -153,7 +155,7 @@ void recording_finalise(void);
 //!            a channel is enabled for recording
 //! \return True if the initialisation was successful, false otherwise
 bool recording_initialize(
-        address_t recording_data_address, uint32_t *recording_flags);
+        void **recording_data_address, uint32_t *recording_flags);
```
alan-stokes: changing void ** to a recording_data_t* and the places around would be a better solution. unfortunately it requires a bit of thought. bounced back at @rowleya

rowleya: OK, so I just went to change this and it reminded me why I didn't do this the first time.  The issue is that the data required is not just a recording_data_t, but also some other things (see the lines below).  So the return is *not* an address to a recording_data_t, so it might be a bit weird to have this type!

alan-stokes: so what ya saying is that were doing really shite things here. and so we need to keep it as a void **

PR #493, opened 2019-09-27
Project eighteen

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNMan/pull/181

Only place to specify n_ cores:
machine.MAX_CORES_PER_CHIP

For max machine used to estimate added a config settig
max_machine_core_reduction = 1

For both max machine and use virtual machines
With with_monitors=True (The default)
The result is one less core

So with the default settings:

actually machine 
50% 17 n_user_processors
50% 16 n_user_processors

Max machine
100% 16 n_user_processors

Virtual machine
100% 17 n_user_processors
 
tested with https://github.com/SpiNNakerManchester/sPyNNaker8/pull/272


Christian-B:
I would even suggest striping the core limit config out and ONLY using Machine.MAX_CORES_PER_CHIP

This could not be set in config as used in several places during import

alan-stokes:
so how you going to supprot tests that want to test with small numbers of cores per chip?

Christian-B:
It will not be possible to support automated tests which want to use a smaller number of chips per core.

There are places where max cores is used to create constants.   

It was not really possible before but it just looked like it was.

alan-stokes:
so you want to remove functionality...... we were able to do this before for VMs. because we tested the spinnaker 2 layout for routing and that worked. I don't like removing that functionality when it can be very very useful for both tests and for spinnaker 2 explorations.  Can we not make it work instead of stripping it out?

Christian-B:
For tests you can still create VirtualMachine of any size that has not changed.

In fact the live code creates VM's with 17 cores for max VM's and 18 for running on VM.s

To set the system wide in the config correctly would require delaying any code that uses max cores until after the config is read in.  Hard and out of scope of this PR

PR #495, opened 2019-09-30
cherry picked app prov folder vs system folder

alan-stokes:
this is the cherry picks from roberts cleaned ear for ensuring iobuf goes to the right folder. app prov vs system prov, which makes debugging much easier as your no longer flooded with iobufs. 

depends upon :
spynnaker https://github.com/SpiNNakerManchester/sPyNNaker/pull/710
spynnaker8 https://github.com/SpiNNakerManchester/sPyNNaker8/pull/275


spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -51,17 +62,27 @@ def __call__(self, graph, placements, graph_mapper=None):
...
                 elif graph_mapper is not None:
-                    machine_verts = graph_mapper.get_machine_vertices(vertex)
+                    machine_verts = graph_mapper.get_machine_vertices(
+                        vertex)
                     for machine_vertex in machine_verts:
                         self._add_vertex_to_subset(
                             machine_vertex, placements,
-                            binary_start_types[placement_binary_start_type])
+                            binary_start_types[
+                                placement_binary_start_type])
```
andrewgait: Don't think this needs splitting across two lines.

unittests/interface/interface_functions/test_front_end_common_chip_iobuf_extractor.py
```diff
@@ -80,6 +81,10 @@ def mock_aplx(id):
 alphaaplx = mock_aplx("alpha")
 executable_targets.add_subsets(alphaaplx, core_subsets)
 
+types = {fooaplx: ExecutableType.USES_SIMULATION_INTERFACE,
+         baraplx: ExecutableType.USES_SIMULATION_INTERFACE,
```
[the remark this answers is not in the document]
alan-stokes: i think thats what @Christian-B intended. i just augmented the test to have the equiv data struct my stuff needed

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -51,17 +62,27 @@ def __call__(self, graph, placements, graph_mapper=None):
...
+                        binary_name)
+                    binary_to_start_type[binary_path] = (
+                        placement_binary_start_type)
 
-        # only got apps with no binary, such as external devices. return no app
+        # only got apps with no binary, such as external devices.
+        # return no app
         if not binary_start_types:
-            return [ExecutableType.NO_APPLICATION]
+            return [ExecutableType.NO_APPLICATION], {}
```
rowleya: Is this check needed?  Doesn't the line below just return an empty dict in this case?

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -303,6 +304,19 @@ def convert_vertices_to_core_subset(vertices, placements):
     return core_subsets
 
 
+def find_executable_start_type(machine_vertex, graph_mapper=None):
+    has_binary = isinstance(machine_vertex, AbstractHasAssociatedBinary)
+
+    if not has_binary:
+        return None
+    elif graph_mapper is not None:
```
rowleya: else not needed...

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -303,6 +304,19 @@ def convert_vertices_to_core_subset(vertices, placements):
...
+        app_vertex = graph_mapper.get_application_vertex(machine_vertex)
+        if isinstance(app_vertex, AbstractHasAssociatedBinary):
+            return app_vertex.get_binary_start_type()
```
rowleya: What do you do if it isn't an instance?  No return in that case... but then you checked this above so maybe not needed here?

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -303,6 +304,19 @@ def convert_vertices_to_core_subset(vertices, placements):
...
+    else:
```
rowleya: Else not needed if you fix the return above

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -853,6 +861,7 @@ def _run(self, run_time, run_until_complete=False):
         # Work out the maximum run duration given all recordings
         if self._max_run_time_steps is None:
             self._max_run_time_steps = self._deduce_data_n_timesteps()
+        clear_injectables()
```
rowleya: As above...

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2294,6 +2308,12 @@ def get_current_time(self):
     def get_generated_output(self, name_of_variable):
         if name_of_variable in self._last_run_outputs:
             return self._last_run_outputs[name_of_variable]
+        if name_of_variable in self._mapping_outputs:
+            return self._mapping_outputs[name_of_variable]
+        if name_of_variable in self._load_outputs:
+            return self._load_outputs[name_of_variable]
+        if name_of_variable in self._machine_outputs:
+            return self._machine_outputs[name_of_variable]
```
rowleya: I don't see what uses this now.  Is there a reason for adding these cases?

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -21,27 +21,38 @@
...
         progress = ProgressBar(
             graph.n_vertices, "Finding executable start types")
         for vertex in progress.over(graph.vertices):
             # try to locate binary type, but possible it doesn't have one
             placement_binary_start_type = None
+            binary_name = None
             if isinstance(vertex, AbstractHasAssociatedBinary):
-                placement_binary_start_type = vertex.get_binary_start_type()
+                placement_binary_start_type = (
+                    vertex.get_binary_start_type())
```
rowleya: Why is this now split when it worked not-split before?

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -21,27 +21,38 @@
...
+                binary_name = vertex.get_binary_file_name()
             elif graph_mapper is not None:
-                associated_vertex = graph_mapper.get_application_vertex(vertex)
-                if isinstance(associated_vertex, AbstractHasAssociatedBinary):
+                associated_vertex = (
+                    graph_mapper.get_application_vertex(vertex))
+                if isinstance(
+                        associated_vertex, AbstractHasAssociatedBinary):
```
rowleya: Why split when not-split before and not changed?

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -21,27 +21,38 @@
...
                     placement_binary_start_type = \
                         associated_vertex.get_binary_start_type()
+                    binary_name = associated_vertex.get_binary_file_name()
 
             # check for vertices with no associated binary, if so, ignore
             if placement_binary_start_type is not None:
-                # update core subset with location of the vertex on the machine
+                # update core subset with location of the vertex on the
+                # machine
```
rowleya: Why split when not changed?

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
```diff
@@ -51,17 +62,27 @@ def __call__(self, graph, placements, graph_mapper=None):
                         vertex, placements,
                         binary_start_types[placement_binary_start_type])
                 elif graph_mapper is not None:
-                    machine_verts = graph_mapper.get_machine_vertices(vertex)
+                    machine_verts = graph_mapper.get_machine_vertices(
+                        vertex)
```
rowleya: Split not needed

spinn_front_end_common/mapping_algorithms/on_chip_router_table_compression/mundy_on_chip_router_compression.py
```diff
@@ -49,15 +49,15 @@ class MundyOnChipRouterCompression(object):
 
     def __call__(
             self, routing_tables, transceiver, machine, app_id,
-            provenance_file_path, compress_only_when_needed=True,
+            system_prov_folder, compress_only_when_needed=True,
```
rowleya: All the changes in this file rename the parameter, but otherwise change nothing.  I don't think the old name was that bad...

spinn_front_end_common/interface/config_handler.py
```diff
@@ -55,6 +55,12 @@ class ConfigHandler(object):
         #
         "_provenance_file_path",
```
Christian-B: If the idea is to split the provenance in two why is the original provenance folder still there?

alan-stokes: becuase there are prov items which are not graph related. aka pacman, router, etc etc

spinn_front_end_common/mapping_algorithms/on_chip_router_table_compression/mundy_on_chip_router_compression.py
alan-stokes: explictness. its not using the prov file path. its specifically using the system one, as its a system operation.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -522,6 +525,10 @@ def set_n_boards_required(self, n_boards_required):
                 "Clash with n_chips_required.")
         self._n_boards_required = n_boards_required
 
+    @property
+    def executable_finder(self):
+        return self._executable_finder
```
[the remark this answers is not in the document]
alan-stokes: used in the ear to add its binary paths to the set. 


        # update finder to look inside ear model binaries location
        globals_variables.get_simulator().executable_finder.add_path(
            os.path.dirname(model_binaries.__file__))

its better as a property to support things like this. the other option is to add it to a __init__ which maybe. dont know.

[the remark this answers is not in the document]
alan-stokes: ok. having looked. will rewire so looks like new model templates. can be removed as a property

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -842,6 +849,7 @@ def _run(self, run_time, run_until_complete=False):
             self._do_mapping(run_time, n_machine_time_steps, total_run_time)
 
         # Check if anything has per-timestep SDRAM usage
+        provide_injectables(self._mapping_outputs)
```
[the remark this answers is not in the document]
alan-stokes: if memory serves, its because the SSA now needs it to meet the criteria properly. am trying to verify now. it may be a change thats really from another cherry pick. i never promsied i got all of these perfect

but its worth noting that this whole injection crud is dodgy, coz it allows something in a completely seperate repo to force this. is probabily safer to wrap everything in ASB with a injectable provider anyhow.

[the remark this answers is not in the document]
alan-stokes: ok. having checked. it seems to be a cherry pick from timing thats slipped into the app folder stuff. as its in both, and i have vague memories of it needing it for the SSA. Ive also ran a few tests on this pr without them in and it seems to work. so removing

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: fixed.

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: the check was already in master. so out of scope of the pr. the change is to make the return value correct in terms of expecting 2 elements and not 1. 

i assume that the check is needed. as without the explicit return, the return would be 2 empty dicts which is not the same as a explicit no application.

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: else definately needed. machine graph levels would return None if that else wasnt there.

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: though i note i dislike the defense. it does have a return. it will fall out of this if set and return None as is the nature of python. Id be willing to put another else in the pile to explciitly return none

this said, the check is needed, the above check was on a machine vertex, not the application vertex, which this check is on.

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: actually. having looked closer. it seems this function isnt called anywhere.  so removing

alan-stokes: actually. its used by the bitfield compressor report. and the appliction chips used. so it is used.

alan-stokes: having relooked. i think i see what you were trying to say. please check new version

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: fixed.

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: fixed

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: fixed. im going to assume somewhere something was set to 79 and so was being flagged incorrectly

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
alan-stokes: fixed

spinn_front_end_common/interface/interface_functions/locate_executable_start_type.py
rowleya: OK, yes see it now

spinn_front_end_common/utilities/helpful_functions.py
rowleya: You misunderstand - the above is a "return", so "elif" can become just "if".  Not critical though...

spinn_front_end_common/utilities/helpful_functions.py
rowleya: OK, I missed the return below somehow...

Christian-B:
Rubbered!

spinn_front_end_common/mapping_algorithms/on_chip_router_table_compression/mundy_on_chip_router_compression.py
rowleya: OK, not that critical

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: I don't see any resolution to this yet...

spinn_front_end_common/utilities/helpful_functions.py
alan-stokes: fixed

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: I would say that if the change isn't needed for the rest of the changes in here, it probably should be removed.  If we really need it, it should probably go in when we do...

alan-stokes: so you want to keep a method buggy?

rowleya: "Buggy"... Well that depends on if you think it is a bug!  I can see what you mean but I think the definition of this is the output from the last call to run (even if the method name isn't explicit).  The other contexts are really for passing from state-to-state than for outside access...

alan-stokes: if ya view is that the name needs changing over actually making it support access to the data whenever. 

then ill remove and add a if for has ran (or whatever the flag is)

PR #496, opened 2019-10-01
Allow for larger recordings

dkfellows:
Ed Jones apparently had more than 1GB of data in one recording region (after quite a pause/resume cycles) and hit a compile-time limit on SQLite (the maximum size of a BLOB data item). Ugh.

----

Yes, we did discuss other ways to reduce the amount of data. We still need to avoid any arbitrary limits here. (There are also limits on the overall size of a database; if we hit those, we might need to think about finding the user their own disks to use for running on as well as fixing our tools.)


dkfellows:
See also SpiNNakerManchester/JavaSpiNNaker#242

dkfellows:
See also SpiNNakerManchester/sPyNNaker8#280

dkfellows:
The good thing about this new implementation is that it avoids concatenating BLOBs with BLOBs in the DB, which is a very slow operation. Python is better at that.

PR #505, opened 2019-10-29
Fix various issues in the docs

dkfellows:
Not deeply fixed, just so that we no longer get warnings.

### DIFF
--- a/spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
+++ b/spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
@@ -660,12 +660,9 @@ class ReverseIPTagMulticastSourceMachineVertex(
             placement, self._REGIONS.RECORDING.value, txrx)
 
     @property
+    @overrides(SendsBuffersFromHostPreBufferedImpl.send_buffers)
     def send_buffers(self):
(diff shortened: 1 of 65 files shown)


spinn_front_end_common/utility_models/reverse_ip_tag_multicast_source_machine_vertex.py
```diff
@@ -644,6 +665,9 @@ def is_recording(self):
     })
     def update_buffer(
             self, first_machine_time_step, run_until_timesteps):
+        """
+        UNUSED?
+        """
```
rowleya: Note the @inject statement - this is used if indirectly.

alan-stokes: fixed in my timing cherry pciks

alan-stokes: i made the buffer updator give the data. instead of it being injected. basically we have an extra api call for buffered input stuff. which avoids this

rowleya: OK, this is merged now - I updated the doc instead for now, but no probs if you have changed it elsewhere.

PR #507, opened 2019-10-31
Max cores per chip

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/124


alan-stokes:
plonking here so that i dont lose track. issues found during recent review. could be in any of the repos. 

1. reudndant brackets, 
2. None issue, 
3. version . height wdith optionals and the code in vm machine generator to go with it, 
4. machine provenance for your logs, 
5. feature chreep in BOARD_48_CHIPS, 
6. magic number in the boot variables.

Christian-B:
-  reudndant brackets,
Fixed

 -  None issue,
better error in VirtualMachine

-  version . height wdith optionals and the code in vm machine generator to go with it,
Leaving width and height in xml as required.  AbstractSpinnakerBase will read these and insert None in not found so there are always values even if None

-  machine provenance for your logs,
provenance not available at the spinnMachine level so went with a report.
As it is a report it is verbose

- feature chreep in BOARD_48_CHIPS,
Lucky you that this useful information is now available.
It stays

- magic number in the boot variables.
I assume this is just in test. so not bothering to remove

PR #509, opened 2019-11-07
check buffer times at app level as machine level does not pick up too long

Christian-B:
import spynnaker8 as sim
sim.setup(timestep=1.0)
input1 = sim.Population(3, sim.SpikeSourceArray(spike_times=[[1], [2], [3], [4]]), label="input1")

This never raises an error as the current check is at the machine level which gets its slice of the full spike times.

requested by @pabogdan

### DIFF
--- a/spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
+++ b/spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
@@ -29,6 +29,7 @@ from .reverse_ip_tag_multicast_source_machine_vertex import (
     ReverseIPTagMulticastSourceMachineVertex)
 from spinn_front_end_common.abstract_models import (
     AbstractGeneratesDataSpecification, AbstractHasAssociatedBinary)
+from spinn_front_end_common.utilities.exceptions import ConfigurationException
 from spinn_front_end_common.utilities.utility_objs import ExecutableType
 from spinn_front_end_common.utilities import globals_variables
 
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
```diff
@@ -146,6 +148,17 @@ def __init__(
         # Keep the vertices for resuming runs
         self._machine_vertices = list()
 
+    def _validate_send_buffer_times(self, n_keys, send_buffer_times):
+        if send_buffer_times is None:
+            return
```
rowleya: Though technically correct, it should may be return None to be explicit.

spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
```diff
@@ -146,6 +148,17 @@ def __init__(
...
+        if len(send_buffer_times) and hasattr(send_buffer_times[0], "__len__"):
+            if len(send_buffer_times) != n_keys:
```
alan-stokes: why use n keys instead of self._n_atoms. i appreciate they're the same value. but n keys is dubious. if it had multiple outgoing partitions, then n keys wouldn't be correct.

Christian-B: Ok changed which also reduces a param for the method.

spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
Christian-B: fixed

PR #512, opened 2019-11-11
Fix to data in protocol

alan-stokes:
this does the following:
1. matches this diagram (which means handling the fail case of when the first packet is lost)
https://docs.google.com/drawings/d/1yT_Vt8QGcyPhY9RJI6lRmhnFyW8N2IsHZgAFv-YP8UI/edit?usp=sharing

2. removed the timing stuff within the SUPG, as this is possibly the reason why things get out of whack, and were not part of the original protocol at all. leaving it in would cause that diagram to cause sooo many other branches which would cause 2 way comms, which the protocol was trying to avoid

jenkins
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/308

 java version here:
https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/243


spinn_front_end_common/interface/spinnaker.cfg
```diff
@@ -228,7 +228,7 @@ n_samples_per_recording_entry = 100
 turn_off_board_after_discovery = False
 
 [Java]
-use_java = False
+use_java = True
```
Christian-B: NO!
We do NOT want Java to be on by Default!

alan-stokes: done. no idea how that slipped

dkfellows:
Branch not deleted: code cleanup in progress. Branch merged because working when `master` bust!

PR #513, opened 2019-11-13
Multirun provenance

rowleya:
Fix for a student having issues with the hard disk filling up.  This is because the current provenance gathering builds up on every run, so you get repeats of the provenance in each file as the runs progress.  This PR fixes this, and also stops the reloading of regions from creating pointless folders.

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -333,9 +333,6 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
 
         "_vertices_or_edges_added",
 
-        # Set of all seen vertext labels
-        "_vertext_labels",
-
         # Version provenance
         "_version_provenance"
     ]
@@ -514,7 +511,6 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
 
         self._last_except_hook = sys.excepthook
         self._vertices_or_edges_added = False
-        self._vertext_labels = set()
 
     def set_n_boards_required(self, n_boards_required):
         """
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -334,7 +334,10 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
         "_vertices_or_edges_added",
 
         # Set of all seen vertext labels
-        "_vertext_labels"
+        "_vertext_labels",
+
+        # Version provenance
+        "_version_provenance"
     ]
 
     def __init__(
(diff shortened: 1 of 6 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -334,7 +334,10 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
         "_vertices_or_edges_added",
 
         # Set of all seen vertext labels
-        "_vertext_labels"
+        "_vertext_labels",
```
andrewgait: This should be "vertex", not "vertext"; but it also doesn't seem to be used anywhere either.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2072,7 +2081,7 @@ def _recover_from_error(self, exception, exc_info, executable_targets):
             # Extract any written provenance data
             try:
                 extracter = PlacementsProvenanceGatherer()
-                extracter(self._txrx, placements, prov_items)
+                prov_items.extend(extracter(self._txrx, placements))
```
andrewgait: "extractor" not "extracter"

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Ah yes, unfortunately my only change is to add a comma, so this was already in there.  I think it gets used by classes in other modules, so changing it in this PR is out-of-scope and in general is more of a pain...

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Fixed.

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: ill just go with oops. as i persume thats me.

alan-stokes:
il review this once ive got my hands out of the protocol c code.

i dont grasp why this is really needed though. as the reason prov grows with runs is becuase prov is suppose to grow with runs. as its for reproducibility. it needs to be showing each atomic step (aka a run). accuming them (which im guessing is what this does) loses that context, and losing a multi-run to be considered equiv to some other multirun. 

if the student is having issues with disks. they can always turn off prov. and maybe this would imply we need yet another report which can be turned on which is this accum equiv report?

spinn_front_end_common/interface/abstract_spinnaker_base.py
andrewgait: I did a search for it elsewhere and didn't find it, which is why I brought it up here.

andrewgait:
I think the point is that in a multi-run context, often the provenance is (almost) identical when moving from one run to the next, so keeping multiple copies of the same thing is somewhat unnecessary.

rowleya:
Yes, the issue is that it writes one file with provenance A then on the next run it writes a second file with provenance A + B and then on the next A + B + C.  This is clearly a bug.  This PR fixes the build-up, so you get a file with A, then a file with B, then a file with C; you still get all the provenance, just not repeatedly.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: True enough - it is now gone

alan-stokes:
ok, if thats the bug. then fine. like i aid. havent reviewed it yet. still arms deep in c code

PR #518, opened 2019-11-16
Fix injection

rowleya:
It was realised some time back that reverse IP tags are not needed to inject spikes if you know which core the RIPTMCS is on (which we do, because we have a database).  Therefore one does not need to reserve a Reverse IP Tag or specify a port to be using injection.

This adds a flag to say we are, so that the database gets written correctly.


alan-stokes:
question. does this make REverse iptags redundant? if so, could we remove them completely? giving us back some dtcm in the monitors.

rowleya:
Reverse IP tags are mostly redundant now in that we no longer really use them.  I guess there is still a case for supporting them though.  I don't think it will save much in the monitors in any case.

PR #520, opened 2019-11-20
should only report finished when you actually finish successfully.

alan-stokes:
as title states. this should remove a variable in debugging on jenkins


rowleya:
I disagree with this; finished means that the code isn't running, not whether it finished cleanly or not.  This would mean that users might end up with a hard disk full of errors that they then have to delete.  Better than this would be to ensure that there is an option that doesn't delete the reports at all e.g. n_reports_kept=None.  This allows better control for those users that don't want to delete reports.

alan-stokes:
how about a errored flag which the cleaner can have a config to remove or not?

rowleya:
Yes, that sounds like it could work.  Config could be a count of kept error items, or just whether to include them in the existing count.

alan-stokes:
hows this then?

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2586,6 +2587,8 @@ def stop(self, turn_off_machine=None,  # pylint: disable=arguments-differ
 
         if exc_info is not None:
             reraise(*exc_info)
+        else:
+            self.write_errored_file()
```
rowleya: Isn't this the wrong way around i.e. it will write the errored file when it *hasn't* got an error here?

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1106,6 +1106,7 @@ def _run_algorithms(
             try:
                 self._shutdown()
                 self.write_finished_file()
+                self.write_errored_file()
```
rowleya: This means you are going to write it when there isn't an error?

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: oops

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: this is inside the except of the run_algorithm. so to even be here, its had some form of error. it may be recoverable to shutdown and "finish" but it finished via a error state. so no it'll write it when there has been an error. unless ive misread the code again

spinn_front_end_common/interface/abstract_spinnaker_base.py
alan-stokes: teach me for using the old version as logic point to not read. fixed

alan-stokes:
reready for ya @rowleya

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Ah OK.  I just noticed, this should also write the finished and errored files outside of the try catch I think (this isn't strictly part of this PR. but easy enough to fix whilst here).  This would mean that it would write the files even if shutdown were to fail.

spinn_front_end_common/interface/config_handler.py
```diff
@@ -329,18 +345,29 @@ def _set_up_output_folders(self, n_calls_to_run):
...
+        """ Writes a errored file that allows file removal to only remove \
+            folders that are errored if requested to do so
+        :rtype:
+        """
+        self.__write_named_file(ERRORED_FILENAME)
 
     def set_up_timings(self, machine_time_step=None, time_scale_factor=None):
         """ Set up timings of the machine
-
+
```
rowleya, line 370:
```suggestion

```

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1169,10 +1169,11 @@ def _run_algorithms(
                     exc_info=True)
             try:
                 self._shutdown()
+                self.write_finished_file()
+                self.write_errored_file()
             except Exception as e3:
                 logger.warning("problem when shutting down {}".format(e3),
                                exc_info=True)
```
rowleya, line 1176:
```suggestion
            except Exception as e3:
                logger.warning("problem when shutting down {}".format(e3),
                               exc_info=True)
            self.write_errored_file()
            raise e
```

rowleya (changes requested):

spinn_front_end_common/interface/config_handler.py
```diff
@@ -329,18 +345,28 @@ def _set_up_output_folders(self, n_calls_to_run):
...
+    def write_errored_file(self):
+        """ Writes a errored file that allows file removal to only remove \
+            folders that are errored if requested to do so
+        :rtype:
+        """
+        self.__write_named_file(ERRORED_FILENAME)
 
     def set_up_timings(self, machine_time_step=None, time_scale_factor=None):
         """ Set up timings of the machine
```
rowleya, line 369:
```suggestion
        """ Set up timings of the machine

```

PR #532, opened 2019-12-05
Fix router table copy

rowleya:
Fixes router table copying to do "the right thing".  Instead of trying to allocate and free the heap, this simply copies the entries from the SARK SDRAM copy directly, overwrites them (using SARK function rtr_mc_set, which also updates the SARK SDRAM copy so ybug rtr_dump still works), and then restores them afterwards as if nothing had happened.

Note that the SARK SDRAM copy is copied directly rather than going through the SARK read function, as the SARK read function strips the app_id, so when the entries are copied back this would be lost.

Note also that this will work provided that no one changes anything on the router heap between the copy being made and the copy being restored.  This is no different than master of course!

And yes, I am aware that the word "copy" has been used many times in the above description!  Sorry!


rowleya:
Currently being tested with SpiNNakerManchester/sPyNNaker8#324

c_common/models/extra_monitor_support/src/extra_monitor_support.c
```diff
@@ -904,52 +905,21 @@ static INT_HANDLER data_in_process_mc_payload_packet(void) {
...
-            // try setting the valid router entry
-            if (rtr_mc_set(idx + start_entry_id, sdram_address[idx].key,
-                    sdram_address[idx].mask, sdram_address[idx].route) != 1) {
-                io_printf(IO_BUF, "Failed to write router entry %d, "
-                        "with key %08x, mask %08x, route %08x\n",
-                        idx + start_entry_id, sdram_address[idx].key,
-                        sdram_address[idx].mask, sdram_address[idx].route);
-            }
-        }
+        rtr_mc_set(idx + N_BASIC_SYSTEM_ROUTER_ENTRIES,
```
alan-stokes: loss safety code for when the set fails

c_common/models/extra_monitor_support/src/extra_monitor_support.c
```diff
@@ -841,7 +841,8 @@ static void data_in_clear_router(void) {
         if (rtr_mc_get(entry_id, &router_entry) &&
                 router_entry.key != INVALID_ROUTER_ENTRY_KEY &&
                 router_entry.mask != INVALID_ROUTER_ENTRY_MASK) {
-            rtr_free(entry_id, 1);
+            rtr_mc_set(entry_id, INVALID_ROUTER_ENTRY_KEY,
```
alan-stokes: this claims ALL the routing table entries. which means if we ever want to add an extra entry, where there python thinks say there's 1022 entries free. it'll go via spinnman to allocate 1, and get told there's none left. coz the rest are all been allocated to be invalid. bad bad bad

c_common/models/extra_monitor_support/src/extra_monitor_support.c
```diff
@@ -904,52 +905,21 @@ static INT_HANDLER data_in_process_mc_payload_packet(void) {
...
-        }
+        rtr_mc_set(idx + N_BASIC_SYSTEM_ROUTER_ENTRIES,
+            sdram_address[idx].key, sdram_address[idx].mask, sdram_address[idx].route);
     }
 }
 
 //! \brief reads in routers entries and places in application sdram location
 static void data_in_save_router(void) {
-    rtr_entry_t router_entry;
+    rtr_entry_t *rtr_copy = sv->rtr_copy;
```
alan-stokes: @lplana requested when we originally did this, that we not go behind the back of the rtr_mc_get etc. so not sure how happy @lplana  will be with this.

c_common/models/extra_monitor_support/src/extra_monitor_support.c
rowleya: Technically this is correct, but nothing gives parameters outside of the range that would result in an error.  If you think this is a problem, the check could be done quite easily.

c_common/models/extra_monitor_support/src/extra_monitor_support.c
rowleya: It is impossible to not go around the back of the API since the API strips a critical bit of information from the result (the app_id).

c_common/models/extra_monitor_support/src/extra_monitor_support.c
rowleya: This is not true; it *overwrites* all routing table entries.  Given the method is called "data_in_clear_router", this is surely the correct thing to do, and nothing different from what "rtr_free" would do, though only if given the correct entry!

alan-stokes: having discussed via skype. im inclined to accept it all. if @lplana has no objections

c_common/models/extra_monitor_support/src/extra_monitor_support.c
alan-stokes: still want the safety check. no harm in it, and it allows us to a level of safety if we buger up other things that exhibit themselves here

PR #535, opened 2020-01-06
Coverage4

Christian-B:
coverage 5 appears broken!


dkfellows:
Fun and games with versioning when idiots parachute breakage into our code! Yay… (not yay)

dkfellows:
Ah! I see! The problem is actually that coverage 5.0.2 has not been tested with `--cov-append`…

PR #538, opened 2020-01-23
avoid passing paths only used in error get from simulator

Christian-B:
Why pass lots of variable in case there is an error.

Get them from where they are stored! the simulator!

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/chip_iobuf_extractor.py
+++ b/spinn_front_end_common/interface/interface_functions/chip_iobuf_extractor.py
@@ -25,6 +25,7 @@ from spinn_front_end_common.utilities.helpful_functions import (
     convert_string_into_chip_and_core_subset)
 from spinn_machine.core_subsets import CoreSubsets
 from spinnman.model.io_buffer import IOBuffer
+from spinn_front_end_common.utilities.globals_variables import get_simulator
 
 logger = FormatAdapter(logging.getLogger(__name__))
 ERROR_ENTRY = re.compile(r"\[ERROR\]\s+\((.*)\):\s+(.*)")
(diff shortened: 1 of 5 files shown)


alan-stokes:
we have a inputs list for the workflow manager for a reason. If your going to be this silly do it for everything. or use the inputs properly.

dkfellows:
> we have a inputs list for the workflow manager for a reason. If your going to be this silly do it for everything. or use the inputs properly.

But this is just like having an environment variable. It isn't something that an execution changes or creates; it's just a thing that exists. You seem to still have an undergraduate's fear of global variables, even for global things…

alan-stokes:
no. id rather have 1 place to find things. instead of spreading things over x global locations. esp as there is a possibility that things will change these "globals" at some point. 

either you have a full global, and you then wreck the entire workflow control system. or you push everything into the work flow manager. like we currently do. some mixture of these is just going to make usage more confusing. which pl;ace do i find x from?

dkfellows:
One way to assuage Alan's concerns would be to pass a callable object that will do any recovery actions needed in general, and that can bind whatever values it needs from wherever as required using whatever mechanism makes sense. The good thing is that this would be far more Pythonic.

spinn_front_end_common/interface/interface_functions/chip_iobuf_extractor.py
```diff
@@ -46,9 +47,20 @@ def __init__(self, recovery_mode=False,
...
-            binary_executable_types, from_cores="ALL", binary_types=None):
+            self, transceiver, executable_targets, executable_finder=None,
+            app_provenance_file_path=None, system_provenance_file_path=None,
+            binary_executable_types=None, from_cores="ALL", binary_types=None):
+
+        sim = get_simulator()
+        if executable_finder is None:
+            executable_finder = sim._executable_finder
+        if app_provenance_file_path is None:
+            app_provenance_file_path = sim._app_provenance_file_path
```
rowleya: This should become public if this is going to be used in this way, as otherwise someone might move it...

spinn_front_end_common/interface/interface_functions/chip_iobuf_extractor.py
```diff
@@ -46,9 +47,20 @@ def __init__(self, recovery_mode=False,
...
+        if system_provenance_file_path is None:
+            system_provenance_file_path = sim._system_provenance_file_path
```
rowleya: Similar issue accessing a private variable

rowleya:
I am happy enough for the simulator to be the object where things come from, but those things that come from it should be "public".

Christian-B:
So out of data easier to start again.

See https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/779

Christian-B:
leaving branch until above issue handled

PR #540, opened 2020-01-30
Fix unaligned recording

dkfellows:
https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/252


andrewgait:
.. apart from that it doesn't compile now, because the makefile log converter only converts *.c files...

dkfellows:
> log_error(...) currently can't be called from a *.h file as it is not converted.

I see. Fixed

PR #542, opened 2020-02-05
Pop based master pop table

rowleya:
This exposes the buffer manager "request_data" method.  This is used by the sPyNNaker pop_based_master_pop_table branch (see PR SpiNNakerManchester/sPyNNaker#776) to read weights, which then avoids needing to pass around a load of variables.


spinn_front_end_common/interface/buffer_management/buffer_manager.py
```diff
@@ -188,9 +188,9 @@ def __init__(self, placements, tags, transceiver, extra_monitor_cores,
                     self._extra_monitor_cores_by_chip,
                     self._packet_gather_cores_to_ethernet_connection_map)
 
-    def _request_data(self, transceiver, placement_x, placement_y, address,
-                      length):
-        """ Uses the extra monitor cores for data extraction.
+    def request_data(
```
alan-stokes: until proved, this being changed to public is a fail

rowleya:
This isn't the right thing to do; using this causes failures!

PR #546, opened 2020-02-13
Provenance sql

Christian-B:
Provence as sql by default!


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2042,6 +2046,12 @@ def _write_provenance(self, provenance_data_items):
             writer = ProvenanceXMLWriter()
         elif self._provenance_format == "json":
             writer = ProvenanceJSONWriter()
+        elif self._provenance_format == "sql":
```
alan-stokes: unless im reading this incorrectly. your prov cuttoff is not accessable. either its xml, json or sql. 

at which point, what value makes you get false to all those 3 and makes you onto the extra else's? 

shouldnt there be a pure IF for the len metric?

Christian-B: There is fourth acceptable value "auto"
See line 588 above.

rowleya: Is this considered ready now?  I can cope with sql being the default...

PR #483, opened 2019-09-04
Graph mapping sanity

dkfellows:
Downstream of SpiNNakerManchester/PACMAN#256

PR #553, opened 2020-02-25
Notification declutter

Christian-B:
Agreed with @dkfellows that is a better name clash fix than what was done in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/483

Removes the name clash and two levels of wrapping.

testing with https://github.com/SpiNNakerManchester/sPyNNaker8/pull/36


spinn_front_end_common/interface/interface_functions/application_runner.py
```diff
@@ -83,7 +61,8 @@ def run_application(
         buffer_manager.load_initial_buffers()
 
         # wait till external app is ready for us to start if required
-        notifier.wait_for_confirmation()
+        if wait_on_confirmation:
```
dkfellows: This test should be moved into real `NotificationProtocol`, as that already has the `wait_on_confirmation` flag (as `self._wait_for_read_confirmation`).

Christian-B: Yes.  So nuked the repeat param in Application Runner/

Added the test to wait_for_confirmation but also to send_read_notification to not save a wait that will never be used.

PR #556, opened 2020-02-27
Revert "Revert "Fix router table copy""

rowleya:
Reinstate this;  it is currently broken so needs fixing first...


Christian-B:
Before merging please test with a spin three board.

For example with fast on run in one go the three tests in 
https://github.com/SpiNNakerManchester/sPyNNaker8/tree/master/p8_integration_tests/quick_test/test_debug_mode

alan-stokes:
??????????? do we need this? is there still a bug?

alan-stokes:
if there is a bug. dont we want to ensure we solve it before the release?

rowleya:
If the existing code is working without issues, it may be worth not bothering with this.

alan-stokes:
given that vague memories says this made bugs, not solved bugs...... i'm inclined to burn it to the ground.

rowleya:
Yes, this version did introduce some bugs.  I don't know if it is worth keeping the work just in case, but the PR can be closed.

rowleya:
Keeping in case it becomes useful

alan-stokes:
question. maybe we can have a project for these kinda things. blue sky prs that we keep around, and then it would be easier in future to see if we ever really do want them again?

rowleya:
Sounds like a good idea

alan-stokes:
cant be. i made it. but fine. will set it up. any other prs you want in there off the top of your head?

rowleya:
Nothing I can think of.  Question; do we want to leave the PRs open?  I guess we can filter out the project...

alan-stokes:
i cant put them in the project if they closed i think.

alan-stokes:
meh. still in project. fine. left now

PR #560, opened 2020-03-04
Better handling of binaries

rowleya:
Fix to not need executables when in virtual mode


rowleya:
Tested by SpiNNakerManchester/sPyNNaker8#369 (mainly the unit tests in this case)

Christian-B:
I would like to see this tested with a cherry pick from https://github.com/SpiNNakerManchester/SpiNNUtils/pull/88

rowleya:
I did think about doing that; the cherry pick was complex because the commits weren't isolated.  I will have a look and see if I can extract the important bit

rowleya:
OK, I have created this PR in SpiNNUtils: SpiNNakerManchester/SpiNNUtils#113.  Will now force-rerun tests...

rowleya:
From what I can see in the unit tests of sPyNNaker8, it works correctly even when the ExecutableFinder raises an Exception when binaries are not found.

PR #561, opened 2020-03-05
Use bulk inserts for faster DB generation

dkfellows:
This uses `insertmany` to do the insertions very quickly and hoists as much of the computation into the DB as possible. It also ensures that we perform the insertions within a single transaction, rather than using one transaction per insertion.


Christian-B:
Looks good to me.

OK to merge when @dkfellows is happy all testing he plans to do is done.

dkfellows:
Just waiting for the integration testing to cook.

PR #565, opened 2020-03-09
Count router errors

rowleya:
This enables the counting of router errors; admittedly it only gets enabled when the extra monitor is enabled, but I think we can live with that restriction...


c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -157,15 +157,21 @@ enum {
     RTR_BLOCKED_BIT = 25,
     RTR_FPE_BIT = 17,
```
lplana: Seems that RTR_FPE_BIT  should be 18, as the number of processor cores.

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -157,15 +157,21 @@ enum {
...
     RTR_LE_BIT = 6,
-    RTR_DENABLE_BIT = 2
+    RTR_PARITY_COUNT_BIT = 5,
+    RTR_FRAME_COUNT_BIT = 4,
+    RTR_TS_COUNT_BIT = 3,
+    RTR_DUMP_ENABLE_BIT = 2
 };
```
lplana: Some of these bits correspond to the router control register and some to the router status register. Would it make the code clearer (less error prone) if different 'enums' were used for bits in different router registers?

Currently, 'spinnaker.h' contains only register addresses and offsets. Should these bit-level definitions move there to have a common source for hardware-related data?

c_common/models/system_models/src/extra_monitor_support.c
rowleya: Oops - fixed it now!

c_common/models/system_models/src/extra_monitor_support.c
rowleya: Yes, I agree that things from different registers should be in different places so I have done this.  The masks are still in a single place so far, as these are more the use of the bits so fairly specific to this code.

I would be happy if spinnaker.h contained these sort of definitions since they are hardware specific.  I guess the masks could also be there in that case.  I don't think this directly affects this PR however; it can be done outside of these changes and then a new PR can be done to make more changes to this file.

lplana: Fully agree.

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: Re-fixed this. (It's buggy on master.)

alan-stokes:
my main question for this is why are we not reporting these as separate drop types in the reinjection status? Id imagine as a reproducability fella id want to know if these hardware faults were the reason. Heck just so we know the hardware is being flaky due to hardware flakes vs random drops.

rowleya:
These errors are currently reported as part of the router provenance.  See here:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/master/spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py#L393-L402

Note that this isn't a counter, just a flag.  It does gather error "types" so you can see what happened but not how often.  Any error here is an indication of a hardware failure though; the software cannot generate an error like this.

alan-stokes:
correct me if i'm wrong. but couldn't we use RTR_ERRCNT_MASK to check on a interrupt by interrupt fashion to count the number of these found? 

given the overall "it found something which should never occur" means we'll flag it up differently, so fine. But out of pure curisoty would we be able to use that to track how many hardware screw ups we had?

rowleya:
Sorry, I have just checked again and the code here enables the counting of total errors.  The code already in master will read the count and store it and report it.  Note that you can't get a count of individual error types (the hardware doesn't support this), just a total error count (which is saturating and resets when we read it during provenance gathering).  So what you get is a list of the errors that have been flagged at some point, and a count of total errors.  This is the best that the hardware can give you.

PR #566, opened 2020-03-11
Documentation for algorithms

dkfellows:
Be aware that this splits up a few bits of code in order to make them easier to apply types to.


spinn_front_end_common/interface/ds/__init__.py
```diff
@@ -16,8 +16,10 @@
...
 from .data_row_reader import DataRowReader
 from .data_row_writer import DataRowWriter
+from .ds_write_info import DsWriteInfo
 
 __all__ = [
     "DataRowReader",
     "DataRowWriter",
-    "DataSpecificationTargets"]
+    "DataSpecificationTargets",
+    "DsWriteInfo"]
```
dkfellows: This line is so that the `DsWriteInfo` class gets documented and linked.

alan-stokes:
this feels like a stupid question. but shouldn't all these docs have text explaining what they are? 

and this needs conflict fixing

PR #568, opened 2020-03-17
End without run bug

Christian-B:
This adds two small fixes.

self._last_run_outputs is only set by run so if it is is None we can assume not run so need for the confusing error.

Append token "ApplicationRun" if not running due to empty graph.
For some reason my system all of a sudden started giving token missing errors when trying to run BufferExtractor.
Unsure why this happened but this fixes it.


rowleya:
Looks fine - I think that adding that the application has run is a good solution for virtual board in any case.

PR #569, opened 2020-03-17
Power provenance

dkfellows:
This adds reporting of power usage via the provenance mechanism, as requested by @pabogdan. Also involved was fixing a number of bugs in the power recording, as the mechanism as described was no longer correct (bulk data transfers are now done using a separate algorithm).


dkfellows:
I'm not sure what integration tests ought to be done for this branch yet.

Christian-B:
ComputeEnergyUsed requires a SpallocServer which means it can not work with either a local board or a HBO connection

This needs fixing!

dkfellows:
That's what I get for copying crap from the original report code.

dkfellows:
The calculations themselves are what they are. I've just tried to make sure that:

1. The measurements are applied to the right things.
2. The results of the measurements are stored in a general description first, and then that's fed to the (vastly simpler!) report generation and provenance generation.

dkfellows:
OK, passing all tests including my new one.

dkfellows:
@pabogdan We now write power information to the provenance system _if_ provenance saving is turned on _and_ the energy report is turned on:
```
[Reports]
write_energy_report = True
write_provenance_data = True
```
The provenance reporting might be on by default. Energy reporting is definitely not as it has noticeable overhead.

pabogdan:
Cool stuff! Now I just need to bring my branches up to date with master. This is definitely not a Friday afternoon job.

@meltingCat, you were interested in energy estimates. This is how you enable them.

PR #573, opened 2020-03-24
Move logic into query

dkfellows:
This avoids a two-stage query for inserting data by moving the lookup of ethernet IDs into the relevant inserts.


dkfellows:
Thank goodness for our integration tests.

PR #574, opened 2020-03-25
Pass LPG Params through all the way to the relevant vertices

dkfellows:
This is *so* much better than having an enormous number of arguments!


dkfellows:
Resolves SpiNNakerManchester/sPyNNaker8#388

PR #579, opened 2020-03-31
Cheaper disable/enable of interrupts in logging

dkfellows:
Because interrupt manipulation *cannot* be optimised out by the compiler (as it doesn't really understand it) the placement of calls outside the check of whether a logging call would be done meant that every log call, even the disabled ones, resulted in two expensive calls being inserted. We really don't want that!

Was caused by insufficient caution in the review of #577


dkfellows:
In theory, the locking ought to be inside SARK because its current implementation only locks when allocating memory.  I guess that the case of having an interrupt (that contains a print) happen during printing didn't occur to the author of the code.

Christian-B:
Looks good to me.

PR #582, opened 2020-04-07
Actually call ChipIOBufClearer when in multirun mode

andrewgait:
Logic seemed wrong here, and it meant that ChipIOBufClearer wasn't being called in multi-runs, so the iobuf was repeating (and getting confusing...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_install:
 install:
   # These things ought to be enough
   - pip install -r requirements-test.txt
-  - pip install pylint python-coveralls 'coverage>=4.4,<5.0'
+  - pip install 'pylint<2.5' python-coveralls 'coverage>=4.4,<5.0'
   - python ./setup.py install
 
 before_script:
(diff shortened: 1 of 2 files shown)


andrewgait:
So it seems that it wasn't actually the logic around the call to the clearer, but that during auto-pause-and-resume, the graph_changed variable was remaining True throughout.  This is clearly wrong so I've edited it as such for subsequent APAR runs... unless there's a better way of doing it?

alan-stokes:
2 things come to light from that change. 

1. redundant brackets over the if. 
2. given your only doing a bool set. id move the bool set down 1 line and remove the if. given its just as costly. 

im still trying to understand why that fixes your issue. so more to follow possibly

alan-stokes:
2 questions / ideas. 
1. we could instead of only putting the iobuf cleaner in after the first run. we could just put it in every run. if there's an order issue, tokens!
2. we could push the idnex down to do run, and do the if where the decision to put the iobuf cleaner into the algorithm list is instead. which then hit what your trying to do.

a side note. is that changing graph changed to false after the first part of auto pause and resume vs when the code is requested to rerun via the self._detect_if_graph_has_changed(True) method........ Im not sure your gaining much. yes the graph hasnt changed since the last subset of run. and if you called that function at the point your setting it to false. it would return false..... but..... semantically i think its a bit odd to play with that variable to get the iobuf clearer........... the more i lookma t it, to protect against treads adding compoenents to the graph. maybe we shouldnt have that bool, but actually ask the vertices and edges every time? dont know

andrewgait:
>     1. we could instead of only putting the iobuf cleaner in after the first run. we could just put it in every run. if there's an order issue, tokens!

Personally this seems like the best option - it shouldn't hurt to have it there on the first run too.  Will see whether this works...

PR #581, opened 2020-04-03
Real bitfields in C

dkfellows:
This uses real C bitfields (**NB:** not the ones that Alan is struggling with! Not a close relation at all!) to make scrobbling around in the SpiNNaker hardware nicer, which reduces the nastiness of the extra monitor quite a bit. C bitfields are not particularly well known because they're rather non-portable, but the ARM ABI locks them down more (definitely little endian!) and as long as all the bitfield-structs are word sized, use uints for all members, and cover all 32 bits, things should be fine.

This does not make using the hardware *much* easier. You still need to read the datasheet if you're going to do very much.

PR #583, opened 2020-04-08
Real bitfields in C

dkfellows:
This uses real C bitfields (NB: not the ones that Alan is struggling with! Not a close relation at all!) to make scrobbling around in the SpiNNaker hardware nicer, which reduces the nastiness of the extra monitor quite a bit. C bitfields are not particularly well known because they're rather non-portable, but the ARM ABI locks them down more (definitely little endian!) and as long as all the bitfield-structs are word sized, use uints for all members, and cover all 32 bits, things should be fine.

This does not make using the hardware *much* easier. You still need to read the datasheet if you're going to do very much.

See also:

* https://community.arm.com/developer/tools-software/tools/f/keil-forum/15065/constant-bit-fields-in-struct
* https://stackoverflow.com/questions/28853742/whats-the-purpose-of-unnamed-bit-field-at-the-end-of-structure
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36750

(Rework of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/581)


dkfellows:
The static assertion stuff is profound magic. It is code that produces a structure definition that includes an array of legal length (1) when the assertion (that the relevant structure is of length 1 word) holds, and which produces an illegal length (-1) when the assertion fails.

There _is_ an official [`static_assert` in C](https://en.cppreference.com/w/c/language/_Static_assert) and GCC supports it, but [Eclipse does not](https://www.eclipse.org/forums/index.php/t/1085780/) ([bug #445297](https://bugs.eclipse.org/bugs/show_bug.cgi?id=445297)) so I've not used it.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+// 4. ARM 968
+
+// No special registers here
+
+// ---------------------------------------------------------------------
+// 5. Vectored Interrupt Controller
+
+typedef void (*vic_interrupt_handler_t) (void);
+
+typedef union {
```
alan-stokes, line 69: comment what this union is for

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    struct {
+        uint watchdog : 1;
```
alan-stokes, line 71: should we be commeting these seperately?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+        uint cc_rx_fr : 1;
+        uint int0 : 1;
+        uint int1 : 1;
+        uint gpio8 : 1;
+        uint gpio9 : 1;
+    };
+    uint value;
+} vic_mask_t;
+
+typedef struct {
```
alan-stokes, line 107: comment what the struct is for

c_common/front_end_common_lib/include/spinn_extra.h
alan-stokes: this covers all structs. as they all open up the question

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    vic_mask_t int_disable;
+    vic_mask_t soft_int_enable;
+    vic_mask_t soft_int_disable;
+    bool protection;
+    const uint _padding[3];
+    vic_interrupt_handler_t vector_address;
+    vic_interrupt_handler_t default_vector_address;
+} vic_control_t;
+
+typedef struct {
```
alan-stokes, line 122: comment what struct is for

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint source : 5;
+    uint enable : 1;
+    uint : 26;
+} vic_vector_control_t;
+
+ASSERT_WORD_SIZED(vic_mask_t);
```
alan-stokes, line 128: comment saying we're ensuring structs are word aligned?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+ASSERT_WORD_SIZED(vic_interrupt_handler_t);
+ASSERT_WORD_SIZED(vic_vector_control_t);
+
+// ---------------------------------------------------------------------
+// 6. Counter/Timer
+
+typedef struct {
```
alan-stokes, line 135: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint size : 1;
+    uint pre_divide : 2;
+    uint : 1;
+    uint interrupt_enable : 1;
+    uint periodic_mode : 1;
+    uint enable : 1;
+    uint : 24;
+} timer_control_t;
+
+enum timer_pre_divide {
```
alan-stokes, line 146: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    TIMER_PRE_DIVIDE_1 = 0,
+    TIMER_PRE_DIVIDE_16 = 1,
+    TIMER_PRE_DIVIDE_256 = 2
+};
+
+typedef struct {
```
alan-stokes, line 152: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint status : 1;
+    uint : 31;
+} timer_interrupt_status_t;
+
+typedef struct {
```
alan-stokes, line 157: comment

alan-stokes: can i propose that all structs are named at top and bottom./ for readibility purposes?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint _dummy;
+} timer_controller_t;
+
+ASSERT_WORD_SIZED(timer_control_t);
+ASSERT_WORD_SIZED(timer_interrupt_status_t);
+
+// ---------------------------------------------------------------------
+// 7. DMA Controller
+
+typedef struct {
```
alan-stokes, line 174: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint : 2;
+    uint direction : 1; // 0 = write to TCM, 1 = write to SDRAM
+    uint crc : 1;
+    uint burst : 3;
+    uint width : 1; // 0 = word, 1 = double-word
+    uint privilege : 1;
+    uint transfer_id : 6;
+} dma_description_t;
+
+typedef struct {
```
alan-stokes, line 186: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint uncommit : 1;
+    uint abort : 1;
+    uint restart : 1;
+    uint clear_done_int : 1;
+    uint clear_timeout_int : 1;
+    uint clear_write_buffer_int : 1;
+    uint : 26;
+} dma_control_t;
+
+typedef struct {
```
alan-stokes, line 196: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint axi_error : 1;
+    uint user_abort : 1;
+    uint soft_reset : 1;
+    uint : 2;
+    uint write_buffer_error : 1;
+    uint : 3;
+    uint processor_id : 8;
+} dma_status_t;
+
+typedef struct {
```
alan-stokes, line 217: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint user_abort_interrupt : 1;
+    uint soft_reset_interrupt : 1;
+    uint unknown1 : 1;
+    uint unknown2 : 1;
+    uint write_buffer_error_interrupt : 1;
+    uint : 10;
+    uint timer : 1;
+} dma_global_control_t;
+
+typedef struct {
```
alan-stokes, line 235: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint _zeroes : 5;
+    uint value : 5;
+    uint : 22;
+} dma_timeout_t;
+
+typedef struct {
```
alan-stokes, line 241: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint enable : 1;
+    uint clear : 1;
+    uint : 30;
+} dma_stats_control_t;
+
+typedef struct {
```
alan-stokes, line 247: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    const uint statistics[8]; // r16-r23
+    const uint _unused3[41];
+    const void *current_sdram_address; // r65
+    const void *current_tcm_address;
+    const dma_description_t current_description; // r67
+    const uint _unused4[29];
+    uint crc_polynomial[32];
+} dma_t;
+
+ASSERT_WORD_SIZED(dma_description_t);
```
alan-stokes, line 269: checks structs in this section are word aligned.

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -153,51 +154,16 @@ enum {
 };
 
 enum {
-    RTR_DOVRFLW_BIT = 30,
-    RTR_BLOCKED_BIT = 25,
-    RTR_FPE_BIT = 17,
-    RTR_LE_BIT = 6,
-    RTR_DENABLE_BIT = 2
+    ROUTER_TIMEOUT_MAX = 0xFF
```
alan-stokes, line 157: hashdefine?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint : 7;
+    uint output_stage : 2;
+    uint : 3;
+    uint interrupt_active_dump : 1;
+    uint interrupt_active_error : 1;
+    uint interrupt_active : 1;
+} router_status_t;
+
+enum output_stage {
+    output_stage_empty,
```
alan-stokes, line 400: cap enum names. for consistency with rest of the tools. (also i was thinking i was looking for a variable for a while when reviewing later files)

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -699,9 +672,9 @@ static inline int reinjection_get_status(sdp_msg_t *msg) {
             (reinjector_status_response_packet_t *) &msg->arg1;
 
     // Put the router timeouts in the packet
-    uint control = (uint) (rtr[RTR_CONTROL] & 0xFFFF0000);
-    data->router_timeout = (control >> 16) & ROUTER_TIMEOUT_MASK;
-    data->router_emergency_timeout = (control >> 24) & ROUTER_TIMEOUT_MASK;
+    router_control_t control = router_control->control;
```
alan-stokes, line 675: this going to also sound stupid. but are they not the wrong way around?
timeout = emergency wait             vs             timeout = wait
emergency timeout =wait.                              emergency timeout =  emerngecy wait

c_common/front_end_common_lib/include/spinn_extra.h
alan-stokes: lost interest. but i believe all these structs and values should have a doc associated with them. likely linking to the datasheet pages

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -325,7 +291,7 @@ enum callback_priorities {
 // ------------------------------------------------------------------------
 
 // The content of the communications controller SAR register
```
alan-stokes, line 293: that your going to set. not the actual contents

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -1638,9 +1645,35 @@ static void data_out_initialise(void) {
...
+        // enable DMA done and error interrupt
+        .transfer_done_interrupt = true,
+        .transfer2_done_interrupt = true,
+        .timeout_interrupt = true,
+        .crc_error_interrupt = true,
+        .tcm_error_interrupt = true,
+        .axi_error_interrupt = true, // SDRAM error
+        .user_abort_interrupt = true,
+        .soft_reset_interrupt = true,
+        .unknown1 = true, // ???
```
alan-stokes, line 1673: whats with the unknowns?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+    uint : 9;
+    uint transfer_done_interrupt : 1;
+    uint transfer2_done_interrupt : 1;
+    uint timeout_interrupt : 1;
+    uint crc_error_interrupt : 1;
+    uint tcm_error_interrupt : 1;
+    uint axi_error_interrupt : 1;
+    uint user_abort_interrupt : 1;
+    uint soft_reset_interrupt : 1;
+    uint unknown1 : 1;
```
alan-stokes, line 228: how about unknown: 2? if they are truely unknown and never set.

c_common/models/system_models/src/extra_monitor_support.c
alan-stokes: why are we setting them to true? what do these do?

alan-stokes:
overall this is a great first step. it makes a lot of magic disappear from system code. but it does basically centerlise a lot of balck magic into 1 file, and that file feels like it should be doced to hell, as its the black magic file. 

minus a place where things are maybe the wrong way around. and some unknowns. i think this would be merge-able once doced

dkfellows:
This PR needs work in general. 
image

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+#include <spinnaker.h>
+#include <stdbool.h>
+#if defined(__GNUC__) && __GNUC__ < 6
+// This particular warning (included in -Wextra) is retarded wrong for client
+// code of this file. Only really a problem on Travis.
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+#endif // __GNUC__
+
+// Generates valid code if the named type is one word long, and invalid code
+// otherwise. It would be simpler if we used _Static_assert, but Eclipse seems
```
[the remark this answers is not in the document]
dkfellows: The problem is that `static_assert()` only compiles correctly when put inside a function (don't know why; that'd seem to be something sensible to have outside a function as it is a compile-time constraint!), and this header file doesn't declare any functions. 

The replacement magic works and lets us ensure that things which are supposed to be word-sized (because they're describing the bit-layout of a register) are actually just that. Failing to have that assertion cost me over a day of debugging time…

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: docuemtned

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: The macro itself is documented as doing that. And it's not _alignment_ but _sizing_.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

dkfellows: Propose away. No.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: documented… and there's a link to the [SpiNNaker datasheet](https://spinnakermanchester.github.io/docs/SpiNN2DataShtV202.pdf) (which is the definitive document here) but it's basically just once or twice for the whole file because I'm a bit lazy and don't want to do a massive cut-n-paste in everywhere. (I also don't think the datasheet has internal anchors, so I can only link to the whole document.)

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: Makes sense. Made so.

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: At the hardware level, they're really `wait0` and `wait1` I think. The `wait0` is the time the hardware will wait between the packet arriving in the router's buffer and it being switched to emergency handling (which SCAMP actually never configures, FWIW). The `wait1` is the time between when the router switches to emergency handling and when the packet is dumped.

Probably. This could be a whole lot clearer. _The datasheet is not especially clear about the naming of things in this area either._

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: I think I'm using the right names from the datasheet now.

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: They say that if the corresponding signal is set in another register, an interrupt should be raised. But those bits in that other register are unassigned. Whee! 🤦 

Removed because then I get (slightly) nicer documentation.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Gone

dkfellows:
Much documenting has now been done. Access to the hardware is _and will remain_ black magic (after all, we're doing big banging on hardware here; real device-driver level evil shenanigans) but at least it should be a great deal more readable black magic than before.

It's a shame I've not got the inclination to rewrite SCAMP and SARK to use this. They'd be a lot easier to understand with it… but that's far more work.

c_common/models/system_models/src/extra_monitor_support.c
rowleya: Yes that sounds right.

dkfellows: Correction, they're `wait1` and `wait2`. It doesn't really matter all that much what the labels are.

c_common/front_end_common_lib/include/spinn_extra.h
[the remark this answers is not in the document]
dkfellows: Correction, `_Static_assert` (note prefix) works fine outside functions from at least GCC 5.2.1 onwards (that's the earliest version I can easily check online), `static_assert` would but we can't include `assert.h` (because spinn_common interferes), and the Eclipse parser throws a wobbly about it and that's insanely distracting. But putting it in an otherwise-vacuous function works very well as an alternative, generating a helpful error message when the Bad Case happens. (In theory I could put all that stuff at the end of the header file… but I prefer keeping such things closer at hand.)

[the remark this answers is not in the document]
dkfellows: BTW, I have tested the assertion code. I'm just not going to commit the test precisely because when it fires off something is truly broken. 😁

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: done

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Capitalised

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: I ought to change the documentation of the Python code to talk to the extra monitor. That's where things really are misleading.

dkfellows: See #624 for my thoughts on that. That PR is _strictly_ downstream of this one.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1106 @@
...
+//
+// ------------------------------------------------------------------------
+
+#ifndef __SPINN_EXTRA_H__
+#define __SPINN_EXTRA_H__
+
+#include <spinnaker.h>
+#include <stdbool.h>
+#if defined(__GNUC__) && __GNUC__ < 6
+// This particular warning (included in -Wextra) is retarded wrong for client
```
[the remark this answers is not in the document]
dkfellows: And I prefer the comment inside the `#if`. I'm looking forward to when I can just delete this little snippet entirely.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    vic_interrupt_handler_t vector_address;
+    //! default vector address register
+    vic_interrupt_handler_t default_vector_address;
+} vic_control_t;
+
+//! VIC individual vector control
+typedef struct {
+    uint source : 5;                //!< interrupt source
+    uint enable : 1;                //!< interrupt enable
+    uint : 26;
```
alan-stokes, line 170: no comment about what 26 is.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+ASSERT_WORD_SIZED(vic_mask_t);
+ASSERT_WORD_SIZED(vic_interrupt_handler_t);
+ASSERT_WORD_SIZED(vic_vector_control_t);
+
+//! VIC registers
+static volatile vic_control_t *const vic_control =
+        (vic_control_t *) VIC_BASE_UNBUF; // NB unbuffered!
+//! VIC interrupt handlers. Array of 32 elements.
+static volatile vic_interrupt_handler_t *const vic_interrupt_vector =
+        (vic_interrupt_handler_t *) (VIC_BASE + 0x100);
```
alan-stokes, line 182: whats this 0x100 come from?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+//! VIC individual interrupt control. Array of 32 elements.
+static volatile vic_vector_control_t *const vic_interrupt_control =
+        (vic_vector_control_t *) (VIC_BASE + 0x200);
```
alan-stokes, line 185: same magic number

c_common/front_end_common_lib/include/spinn_extra.h
alan-stokes: also magic numbers 5 and 1. i was automatically thinking link ids. but def not that here. maybe named constant?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+//! Timer control register
+typedef struct {
+    uint one_shot : 1;          //!< 0 = wrapping mode, 1 = one shot
+    uint size : 1;              //!< 0 = 16 bit, 1 = 32 bit
+    uint pre_divide : 2;        //!< divide input clock (see ::timer_pre_divide)
+    uint : 1;
+    uint interrupt_enable : 1;  //!< enable interrupt (1 = enabled)
+    uint periodic_mode : 1;     //!< 0 = free-running; 1 = periodic
+    uint enable : 1;            //!< enable counter/timer (1 = enabled)
+    uint : 24;
```
alan-stokes, line 203: no comment about what 24. also maybe named constant?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+enum timer_pre_divide {
+    TIMER_PRE_DIVIDE_1 = 0,     //!< Divide by 1
+    TIMER_PRE_DIVIDE_16 = 1,    //!< Divide by 16
+    TIMER_PRE_DIVIDE_256 = 2    //!< Divide by 256
+};
+
+//! Timer interrupt status flag
+typedef struct {
+    uint status : 1;            //!< The flag bit
+    uint : 31;
```
alan-stokes, line 216: comment and named constant?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint load_value;            //!< Load value for Timer
+    const uint current_value;   //!< Current value of Timer
+    timer_control_t control;    //!< Timer control register
+    uint interrupt_clear;       //!< Interrupt clear (any value may be written)
+    //! Timer raw interrupt status
+    const timer_interrupt_status_t raw_interrupt_status;
+    //! Timer masked interrupt status
+    const timer_interrupt_status_t masked_interrupt_status;
+    uint background_load_value; //!< Background load value for Timer
+    uint _dummy;
```
alan-stokes, line 230: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+//!     As a secondary function the DMA controller incorporates a ‘Bridge’
+//!     across which its host ARM968 has direct read and write access to
+//!     System NoC devices, including the SDRAM. The ARM968 can use the
+//!     Bridge whether or not DMA transfers are active.
+//! \{
+
+//! DMA descriptor
+typedef struct {
+    uint _zeroes : 2;       //!< Must be zero
+    uint length_words : 15; //!< length of the DMA transfer, in words
```
alan-stokes, line 264: named constant which can then have a comment explaing what the 15 words are.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint : 2;
+    //! read from or write to system bus, see ::dma_direction_t
+    uint direction : 1;
+    uint crc : 1;           //!< check (read) or generate (write) CRC
+    //! burst length = 2<sup>B</sup>&times;Width, B = 0..4 (i.e max 16)
+    uint burst : 3;
+    uint width : 1;         //!< transfer width, see ::dma_transfer_unit_t
+    uint privilege : 1;     //!< DMA transfer mode is user (0) or privileged (1)
+    uint transfer_id : 6;   //!< software defined transfer ID
```
alan-stokes, line 273: why 6?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+
+//! DMA control register
+typedef struct {
+    uint uncommit : 1;      //!< setting this bit uncommits a queued transfer
+    uint abort : 1;         //!< end current transfer and discard data
+    uint restart : 1;       //!< resume transfer (clears DMA errors)
+    uint clear_done_int : 1;            //!< clear Done interrupt request
+    uint clear_timeout_int : 1;         //!< clear Timeout interrupt request
+    uint clear_write_buffer_int : 1;    //!< clear Write Buffer interrupt request
+    uint : 26;
```
alan-stokes, line 296: named constant?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint tcm_error : 1;     //!< the TCM AHB interface has signalled an error
+    //! the AXI interface (SDRAM) has signalled a transfer error
+    uint axi_error : 1;
+    //! the user has aborted the transfer (via ::dma_control_t::abort)
+    uint user_abort : 1;
+    uint soft_reset : 1;    //!< a soft reset of the DMA controller has happened
+    uint : 2;               // Not allocated
+    uint write_buffer_error : 1;    //!< a buffered write transfer has failed
+    uint : 3;
+    uint processor_id : 8;  //!< hardwired processor ID identifies CPU on chip
```
alan-stokes, line 321: why 8?

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint crc_error : 1;     //!< the calculated and received CRCs differ
+    uint tcm_error : 1;     //!< the TCM AHB interface has signalled an error
+    //! the AXI interface (SDRAM) has signalled a transfer error
+    uint axi_error : 1;
+    //! the user has aborted the transfer (via ::dma_control_t::abort)
+    uint user_abort : 1;
+    uint soft_reset : 1;    //!< a soft reset of the DMA controller has happened
+    uint : 2;               // Not allocated
+    uint write_buffer_error : 1;    //!< a buffered write transfer has failed
+    uint : 3;
```
alan-stokes, line 320: comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+} dma_control_t;
+
+//! DMA status register
+typedef struct {
+    uint transferring : 1;  //!< DMA transfer in progress
+    uint paused : 1;        //!< DMA transfer is PAUSED
+    uint queued : 1;        //!< DMA transfer is queued - registers are full
+    uint write_buffer_full : 1;     //!< write buffer is full
+    uint write_buffer_active : 1;   //!< write buffer is not empty
+    uint : 5;
```
alan-stokes, line 306: named constant and comment

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint timeout_interrupt : 1;     //!< interrupt if ::dma_status_t::timeout set
+    uint crc_error_interrupt : 1;   //!< interrupt if ::dma_status_t::crc_error set
+    uint tcm_error_interrupt : 1;   //!< interrupt if ::dma_status_t::tcm_error set
+    uint axi_error_interrupt : 1;   //!< interrupt if ::dma_status_t::axi_error set
+    uint user_abort_interrupt : 1;  //!< interrupt if ::dma_status_t::user_abort set
+    uint soft_reset_interrupt : 1;  //!< interrupt if ::dma_status_t::soft_reset set
+    uint : 2;                       // Not allocated
+    //! interrupt if ::dma_status_t::write_buffer_error set
+    uint write_buffer_error_interrupt : 1;
+    uint : 10;
```
alan-stokes, line 341: commentand named constant

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint timer : 1;         //!< system-wide slow timer status and clear
+} dma_global_control_t;
+
+//! DMA timeout register
+typedef struct {
+    uint _zeroes : 5;       //!< Must be zero
+    uint value : 5;         //!< The timeout
+    uint : 22;
```
alan-stokes, line 349: comment and named constant

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: The numbers here are bit widths of fields. **I'm _not_ naming those** because that would significantly obscure what the code is doing. C structure fields with explicit bit widths are unusual in most C programming as they're not portable, but they're fine here because we've got a locked down architecture and all supported compilers treat it the same way. _But I want them to look like exactly what they are_ so that a general reader of the C code knows exactly what trickery is going on.

The point of this branch is that it uses this sort of trickery here to turn the _very complicated_ mess that is our hardware interface into a collection of C structures that are significantly easier to use. The use of bit fields is critical to this; they enable getting rid of a lot of explicit shifting and masking (as we will have provided enough information for the compiler to generate that for us) and instead we get actual names for things.

rowleya: To add to this: naming them is explicitly a bad idea as it then lets the user access them.  The user shouldn't be able to do this!  I also think that not commenting is also fine here as otherwise it makes it harder to read.  This should be a convention; if it isn't labelled it is padding (we could make the convention clear but I doubt anyone would read it).

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: `VIC_ADDR` in `spinnaker.h` (i.e., from the datasheet) except that value is broke given what we're doing it. Using the memory offset from the datasheet directly works better.

I don't know why the VIC uses such a large offset.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Same source.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: I've now noted that all these are `// padding`. (I note in passing that the timers are an ARM part that we just integrate. Their layout. Their rules. This whole file is really about describing what the hardware's registers are.)

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Commented as `padding`. Will not name the constant; it's literally just 31 bits of padding that should be ignored.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: The `24` has no relation to anything else in any part of the program (other than that the total bits in the containing structure add up to 32 because that's the number of bits in a word) and it will not be a value that the rest of our applications have to match. In fact, the compiler uses the values to generate bit masks and shifts and that gets rid of magical constants elsewhere (by replacing them with accesses of fields of structures); using separate constants for this will make the code harder to understand.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: No.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: See page 24 of the datasheet.
image
It's six because the datasheet says it is six.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: No

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Because that's the width of the field.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Will not use a named constant.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    sc_pll_control_t pll1_freq_control; //!< PLL1 frequency control
+    sc_pll_control_t pll2_freq_control; //!< PLL2 frequency control
+    uint set_flags;             //!< Set flags register
+    uint reset_flags;           //!< Reset flags register
+    sc_clock_mux_t clock_mux_control;   //!< Clock multiplexer controls
+    const sc_sleep_status_t cpu_sleep;  //!< CPU sleep (awaiting interrupt) status
+    sc_temperature_t temperature[3];    //!< Temperature sensor registers [2:0]
+    const uint _padding[3];
+    //! Read sensitive semaphores to determine MP
+    const sc_mutex_bit_t monitor_arbiter[32];
```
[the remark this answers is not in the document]
dkfellows: No, because they're not related. There's no reason for the number of arbiter bits to be the same as the number of test-and-set and test-and-clear registers. (The latter two are linked because of the way the system controller works.)

c_common/models/chip_power_monitor/src/chip_power_monitor.c
```diff
@@ -134,7 +135,7 @@ static void resume_callback(void) {
 //! The counter for the core is incremented if the core is active.
 //!
 //! Uses get_sample() to obtain the state of the cores.
-static void count_core_states(void) {
+static inline void count_core_states(void) {
```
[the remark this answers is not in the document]
dkfellows: The code is _adding_ the `inline`…

c_common/models/chip_power_monitor/src/chip_power_monitor.c
```diff
@@ -90,27 +91,27 @@ static uint32_t sample_frequency;
...
 //! \return The number of times a busy loop must run.
-static uint32_t get_random_busy(void) {
+static inline uint32_t get_random_busy(void) {
     return (spin1_rand() >> 4) & ((1 << NUM_RANDOM_BITS) - 1);
 }
 
 //! \brief Synchronously records the current contents of the core_counters to
 //! the recording region.
-static void record_aggregate_sample(void) {
+static inline void record_aggregate_sample(void) {
```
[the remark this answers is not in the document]
dkfellows: The code is _adding_ the `inline`…

c_common/models/chip_power_monitor/src/chip_power_monitor.c
```diff
@@ -90,27 +91,27 @@ static uint32_t sample_frequency;
 //!
 //! Note that this accesses into the SpiNNaker System Controller hardware (see
 //! Data Sheet, section 14, register 25).
-static uint32_t get_sample(void) {
-    return sc[SC_SLEEP] & ((1 << NUM_CPUS) - 1);
+static inline uint32_t get_sample(void) {
```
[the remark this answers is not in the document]
dkfellows: The code is _adding_ the `inline`…

c_common/models/chip_power_monitor/src/chip_power_monitor.c
```diff
@@ -90,27 +91,27 @@ static uint32_t sample_frequency;
...
+    return system_control->cpu_sleep.status;
 }
 
 //! \brief Computes a random value used to break up chance periodicities in
 //! sampling.
 //! \return The number of times a busy loop must run.
-static uint32_t get_random_busy(void) {
+static inline uint32_t get_random_busy(void) {
```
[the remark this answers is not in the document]
dkfellows: The code is _adding_ the `inline`…

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -609,25 +530,49 @@ static ushort my_addr;
 //! The SARK virtual processor information table in SRAM.
 static vcpu_t *const _sark_virtual_processor_info = (vcpu_t *) SV_VCPU;
 
-//! \brief DSG metadata
+//! The magic number that marks a valid DSE metadata descriptor
+#define DSE_MAGIC       0xAD130AD6
```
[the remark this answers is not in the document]
dkfellows: Yes, it most certainly is, but the data spec implementation uses spin1_api (for logging) and the extra monitor doesn't (because it needs low-level interrupt access).

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -609,25 +530,49 @@ static ushort my_addr;
...
-    void *regions[];            //!< Pointers to DSG regions
-} dsg_header_t;
+typedef struct dse_header_t {
+    uint magic_number;  //!< Magic number (== #DSE_MAGIC)
+    uint version;       //!< Version (== #DSE_VERSION)
+    void *regions[];    //!< Pointers to DSG regions
+} dse_header_t;
+
+//! Checks that our region data really is region data.
+static void dse_validate(void) {
```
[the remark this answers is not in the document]
dkfellows: In this case, no, there isn't a way to do it. This is due to the peculiarities of the extra monitor.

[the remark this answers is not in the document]
dkfellows: (Or rather, there _is_ a way but it makes a lot of code a lot uglier. Duplicating a small amount of code that's simple enough that we can check it for correctness by eyeball is much cleaner.)

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -609,25 +530,49 @@ static ushort my_addr;
...
 
 //! \brief Get the DSG region with the given index.
 //!
 //! Does *not* validate the DSG header!
 //!
 //! \param[in] index: The index into the region table.
 //! \return The address of the region
-static inline void *dsg_block(uint index) {
-    dsg_header_t *dsg_header = (dsg_header_t *)
+static inline void *dse_block(uint index) {
```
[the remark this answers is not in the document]
dkfellows: For _anything_ except the extra monitor, yes.

dkfellows:
In general, the duplication of data spec stuff in the extra monitor code is because the standard data spec region access implementation depends on spin1_api (mostly for logging when things go wrong) and spin1_api is _forbidden_ to the extra monitor because spin1_api and the extra monitor intercept the same interrupts in incompatible ways. Yes, this is a weirdness of the extra monitor. Duplicating a very short amount of code is way simpler to read than the contortions required to make the data spec region access code work in the two cases.

And I added `inline` in a bunch of places. Stuff was _probably_ being inlined anyway.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+ASSERT_WORD_SIZED(router_timing_counter_ctrl_t);
+ASSERT_WORD_SIZED(router_diversion_t);
+ASSERT_WORD_SIZED(router_fixed_route_routing_t);
+ASSERT_WORD_SIZED(router_diagnostic_filter_t);
+
+//! Router controller registers
+static volatile router_t *const router_control = (router_t *) RTR_BASE;
+//! Router diagnostic filters
+static volatile router_diagnostic_filter_t *const router_diagnostic_filter =
+        (router_diagnostic_filter_t *) (RTR_BASE + 0x200);
```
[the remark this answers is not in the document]
dkfellows: That's a value from the datasheet. I would have loved to have pulled it from spinnaker.h, but temples made that file do weird things with many of its constants.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1439 @@
...
+    uint clock_gen : 1;         //!< System AHB & Clock Gen (pulse reset only)
+    uint entire_chip : 1;       //!< Entire chip (pulse reset only)
+    uint : 2;
+    uint security_code : 12;    //!< ::SYSTEM_CONTROLLER_MAGIC_NUMBER to enable write
+} sc_magic_subsystem_map_t;
+
+//! System controller last reset status
+typedef struct {
+    uint reset_code : 3;        //! One of ::sc_reset_codes
+    uint : 29;
```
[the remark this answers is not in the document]
dkfellows: It's a field width. I'm deliberately not making those named constants so that they're more obviously field widths.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: (I can probably guess the immediate reason, but why they — ARM — had to use such smartass trickery in the address decode logic in the VIC I don't know. It is what it is.)

dkfellows:
Also, there are plenty of magic numbers about. Access to the hardware is indeed all about magic numbers, and always was; the majority of the SpiNNaker datasheet is about defining what the magic numbers are. The purpose of `spinn_extra.h` is to remove magic numbers from _elsewhere_ where possible, but there was, is, and always will be magic about.

dkfellows:
(A few of the more magical numbers and offsets were determined entirely by examining the source to SCAMP. This is particularly an issue with the SDRAM and Ethernet interfaces, but touching anything to do with GPIO/JTAG is also very scary stuff.)

dkfellows:
Thinking about it, using `NUM_LINKS` and `NUM_CPUS` when the field width is determined by the number of links or cores (respectively) makes sense. There are still quite a few fields that are 6 wide for other reasons and I don't plan to cut those over; e.g., PLL clock multipliers have nothing to do with the link count.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: commented

dkfellows:
The reasoning for whether to use before (`//! ...`) or after (`//!< ...`) comments is my own and simple.

All multiline comments categorically go before. All top-level declarations (functions, types, enums, #defines) have comments before. (Functions and complex defines probably need multiline comments anyway.) Before-comments are always indented the same as the thing which they are documenting.

Structure _members_ and enum _members_ have comments after _if they fit on the same line_ with a mostly consistent indentation, since **otherwise the structs get so blooming visually long** on the page. I generally prefer those after/inline comments to begin after 32 characters and with that as a multiple of four (for easy indenting using my <kbd>Tab</kbd> key), but I'm not at all religious about that.

So yes, the decision to use “inline“ comments is purely a matter of [æsthetics](https://dictionary.cambridge.org/dictionary/english/aesthetics). Occasionally, I might break my own rules to make things look nicer, but that's because the _real_ rule is that things have to look good.

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: I wish I could easily have totally unnamed fields for some of our other padding (where it's the size of a word or more) but the contortions for that are too horrible so they just get called things like `_padding3`. If that sort of name doesn't warn someone off, nothing will.

dkfellows:
Turns out that the code using `sark_delay_us()` really wasn't working, and was causing four integration tests to fail. I'm guessing that the issue is the delays were ending up far too long. Still, it wasn't a critical part of this branch.

dkfellows:
Has passed integration testing.

dkfellows:
General design note: `spinn_extra.h` has _lots_ of magic numbers in it because this is where the magic numbers live. Below this is only `spinnaker.h`, which doesn't have many useful magic numbers (and which this file is meant to tame). I'm of the opinion that this file represents a lot of what should have been in `spinnaker.h`, but that file instead uses horrible offsets into a collection of `uint[]`s and that's just nasty and obscure.

The intent is that `spinn_extra.h` will allows us to get rid of magic numbers (for field extraction and packing and so on) from elsewhere.

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -148,14 +137,17 @@ typedef enum data_out_sdp_commands {
 #define TDMA_WAIT_PERIOD   0
 
 //! The initial timeout of the router
-#define ROUTER_INITIAL_TIMEOUT 0x004f0000
+#define ROUTER_INITIAL_TIMEOUT 0x4f
```
alan-stokes, line 140: is that actually correct? given the number has changed. just in case its working, but working for the wrong reasons

c_common/models/system_models/src/extra_monitor_support.c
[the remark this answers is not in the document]
alan-stokes: having reviewed this again. i laded on this AGAIN, and i still think (and @rowleya also thinks) that we shouldn't be bringing this in as a clone. we should adjust the dsg code accordingly. either by not using logs in it, or having 2 version which have the different types of print so you can import the right one. ill leave the other pointers of this block resolved, but the overarching issue still needs thought.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+enum sc_frequency_range {
+    FREQ_25_50,                 //!< 25-50 MHz
+    FREQ_50_100,                //!< 50-100 MHz
+    FREQ_100_200,               //!< 100-200 MHz
+    FREQ_200_400                //!< 200-400 MHz
+};
+
+//! System controller clock multiplexing control
+typedef struct {
+    uint pa : 2;                //!< clock selector for A CPUs (1 2 4 7 8 11 13 14 16);
```
alan-stokes, line 1166: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+                                //!< see ::sc_clock_source
+    uint adiv : 2;              //!< divide CPU clock A by Adiv+1 (= 1-4)
+    uint : 1;                   // padding
+    uint pb : 2;                //!< clock selector for B CPUs (0 3 5 6 9 10 12 15 17);
```
alan-stokes, line 1170: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+                                //!< see ::sc_clock_source
+    uint bdiv : 2;              //!< divide CPU clock B by Bdiv+1 (= 1-4)
+    uint : 1;                   // padding
+    uint mem : 2;               //!< clock selector for SDRAM;
```
alan-stokes, line 1174: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+                                //!< see ::sc_clock_source
+    uint mdiv : 2;              //!< divide SDRAM clock by Mdiv+1 (= 1-4)
+    uint : 1;                   // padding
+    uint rtr : 2;               //!< clock selector for Router;
```
alan-stokes, line 1178: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+                                //!< see ::sc_clock_source
+    uint rdiv : 2;              //!< divide Router clock by Rdiv+1 (= 1-4)
+    uint : 1;                   // padding
+    uint sys : 2;               //!< clock selector for System AHB components;
```
alan-stokes, line 1182: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+//! The possible values of a P2P route
+typedef enum {
+    ROUTER_P2P_ROUTE_E,             //!< Route east
+    ROUTER_P2P_ROUTE_NE,            //!< Route north-east
+    ROUTER_P2P_ROUTE_N,             //!< Route north
+    ROUTER_P2P_ROUTE_W,             //!< Route west
+    ROUTER_P2P_ROUTE_SW,            //!< Route south-west
+    ROUTER_P2P_ROUTE_S,             //!< Route south
+    ROUTER_P2P_ROUTE_DROP,          //!< Drop packet
+    ROUTER_P2P_ROUTE_MONITOR        //!< Send to monitor (as determined by
```
alan-stokes, line 757: move comment to line above def, for doc consistency

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+
+// No registers
+
+// ---------------------------------------------------------------------
+//! \name 13. SDRAM interface
+//! \brief The SDRAM interface connects the System NoC to an off-chip SDRAM
+//!     device. It is the ARM PL340, described in ARM document DDI 0331D.
+//! \details Only meaningfully usable by the monitor core when initialising the
+//!     overall chip. Use at other times is very much not recommended.
+//! \warning Do not use these without talking to Luis first!
```
alan-stokes, line 827: what happens if luis no longer works here? also, or what? ya going to steal my bacon supply?

alan-stokes (changes requested):

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -0,0 +1,1525 @@
...
+    uint : 20;                  // padding
+} sdram_status_t;
+
+//! Memory controller command
+typedef struct {
+    uint command : 3;           //!< one of ::sdram_command
+} sdram_command_t;
+
+//! \brief Memory controller commands, for ::sdram_command_t::command
+//! \todo Verify ::SDRAM_CTL_SLEEP, ::SDRAM_CTL_WAKE, ::SDRAM_CTL_ACTIVE_PAUSE
```
alan-stokes, line 847: did you do that? is it left for future? or is it a self reminder thats been left over?

dkfellows: This is stuff that long-term we'll need to check with Luis. It's not pressing because current users of this include file are _not_ to touch the SDRAM controller! (It's done for completeness.)

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Or else we'll accuse you of Dave-level low level hackery. You have been warned!

c_common/front_end_common_lib/include/spinn_extra.h
alan-stokes: fair enough

c_common/front_end_common_lib/include/spinn_extra.h
alan-stokes: this does not seem to be a costly threat. my point was that the comment should explain WHY to talk to luis first.

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: It's the initial value, not necessarily the best option. (The change of the value here is because it has been disaggregated from the containing structure. And I can never decode those weird 8-bit float things in my head anyway.)

alan-stokes: fiar enough then

dkfellows:
Behold, by the power of regular expression search-and-replace!

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: The point is I know that Luis will tell anyone thinking of messing with that to not do it. 😉

alan-stokes: fine, not caring enough to make it more clear.

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -709,69 +956,106 @@ typedef struct {
...
+    ROUTER_P2P_ROUTE_W,
+    //! Route south-west
+    ROUTER_P2P_ROUTE_SW,
+    //! Route south
+    ROUTER_P2P_ROUTE_S,
+    //! Drop packet
+    ROUTER_P2P_ROUTE_DROP,
+    //! Send to monitor (as determined by
+    ROUTER_P2P_ROUTE_MONITOR
+                                    //! ::router_control_t::monitor_processor)
```
alan-stokes, line 1032: power fail

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -1065,166 +1464,254 @@ static volatile sdram_dll_t *const sdram_dll_control =
...
+    // padding
+    uint : 1;
+
+    //! clock selector for SDRAM; see ::sc_clock_source
+    uint mem : 2;
+    //! divide SDRAM clock by Mdiv+1 (= 1-4)
+    uint mdiv : 2;
+    // padding
+    uint : 1;
+
```
alan-stokes, line 1638: gray skull

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -1065,166 +1464,254 @@ static volatile sdram_dll_t *const sdram_dll_control =
...
+    //! clock selector for System AHB components; see ::sc_clock_source
+    uint sys : 2;
+    //! divide System AHB clock by Sdiv+1 (= 1-4)
+    uint sdiv : 2;
+    // padding
+    uint : 7;
+
```
alan-stokes, line 1652: gray skull

c_common/front_end_common_lib/include/spinn_extra.h
dkfellows: Ah. Missed a spot (that was a doc bug)

c_common/front_end_common_lib/include/spinn_extra.h
```diff
@@ -1065,166 +1464,254 @@ static volatile sdram_dll_t *const sdram_dll_control =
...
-    uint sdiv : 2;              //!< divide System AHB clock by Sdiv+1 (= 1-4)
-    uint : 7;                   // padding
-    uint invert_b : 1;          //!< invert CPU clock B
+    //! clock selector for A CPUs (1 2 4 7 8 11 13 14 16); see ::sc_clock_source
+    uint pa : 2;
+    //! divide CPU clock A by Adiv+1 (= 1-4)
+    uint adiv : 2;
+    // padding
+    uint : 1;
+
```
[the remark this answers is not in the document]
dkfellows: Blank line removed

dkfellows:
The remaining issue that you mentioned is something that should be its own PR.

dkfellows:
Only remaining bits are either unimportant or are things that ought to be fixed in another PR and not this one.

PR #584, opened 2020-04-16
Doxygen

dkfellows:
This branch enables using [doxygen](http://www.doxygen.nl/) for generating documentation for our C code, and turns on running it during our build. It also fixes our code so that it actually generates moderately sensible documentation (warning free except with 1.8.17, which has an annoying bug with misinterpreting `static` as a type modifier).

To read the generated docs:
1. install doxygen
2. check out this branch
3. change to the `c_models` directory
4. run doxygen (no arguments needed)
5. open the file `doc/html/index.html` in your browser

Eclipse users: install the _eclox_ plugin from the Marketplace. Nice integration!


dkfellows:
Files added from #436 still need some work

PR #587, opened 2020-04-27
cherry pick malloc extras

alan-stokes:
this is the code used to project mallocs and frees and allows users to verify their pointer / malloc logic is solid. 

WAS A BILZILLION TIMES usful debugging the compressor and sorter. 

its stand alone, so this should cause no ISSUES in relation to master.


alan-stokes:
boo ya!

alan-stokes:
obsoleted by all_algorithms which used it and is in. killing this

PR #592, opened 2020-05-04
With christian table generator

alan-stokes:
reviewed with doc changes and wee fixes to race conditions


alan-stokes:
ermmmm......how was this pointing at master and not killed?

killing!

PR #595, opened 2020-05-07
Don't clear out the iobuf during initialisation steps

andrewgait:
Oops - previous fix to this was over-zealous and cleared the initial iobuf (i.e. anything written prior to the first timer tick) out.  This should fix it. 

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/400

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1974,7 +1974,8 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
             algorithms.append("SdramUsageReportPerChip")
 
         # Clear iobuf from machine
-        if (not self._use_virtual_board and not self._empty_graphs and
+        if (self._has_ran and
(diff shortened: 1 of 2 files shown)


alan-stokes:
im assuming that means the init. quite confused how it would achieve that. unless the iobuf clearer was called between app binaries loading and the app runner..... which then begs the question i hinted at on the orginal vesion. 

TOKENS! coz if im right, then that has_ran means itll by chance run the clear between those points on the second run, aka before app runner. whereas if we put tokens on the iobuf clearer to say it needs to wait till app runner has finished, you coudl remove the bool check and then its handled in a more ........ intuitive position in the flow. 

and the app runner already has a token that it generates
            <token>ApplicationRun</token>

so all youd need to do is add that as a required input to the iobuf clearer in the xml. 

what ya think?

andrewgait:
> unless the iobuf clearer was called between app binaries loading and the app runner

Yes, this is what currently happens.  I'll add the ApplicationRun token and see what happens.

alan-stokes:
if passes. then i approve the changes.

PR #597, opened 2020-05-12
Report overruns correctly

rowleya:
Changes the code so that it reports when overruns are not 0 rather than > 4.  We used to do > 4 as we didn't realise that the code "overran" at the end of simulation whilst writing to SDRAM.  It doesn't do this any more, as the timer is disabled early, so we should report the overruns correctly too.


alan-stokes:
fails push?! how! 

anyhow, im also fixing it in my prov component for filtering packets at end of timer tick. so if we dont fix this here, itll be fixed there

rowleya:
Push failed due to flake8 new version (and historically incorrect command line on our part).  Fixed now I hope...

PR #598, opened 2020-05-13
Make IOBuf clearer work with run_until_complete

andrewgait:
Another iobuf issue, this time noticed by @lplana in PDP2.  This fixes the issue with the ChipIOBufClearer when using run_until_complete(...) - I think it also works properly with run forever (i.e. run(None)) but it would be good if there are more tests for this other than a single GFE example...

Tested by https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/143 and https://github.com/SpiNNakerManchester/sPyNNaker8/pull/402

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1974,7 +1974,8 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
             algorithms.append("SdramUsageReportPerChip")
 
         # Clear iobuf from machine
-        if (not self._use_virtual_board and not self._empty_graphs and
+        if (not run_until_complete and
+            not self._use_virtual_board and not self._empty_graphs and
                 self._config.getboolean("Reports", "clear_iobuf_during_run")):
             algorithms.append("ChipIOBufClearer")
 
@@ -2034,8 +2035,7 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
         if (self._config.getboolean("Reports", "extract_iobuf") and
                 self._config.getboolean(
                     "Reports", "extract_iobuf_during_run") and
-                not self._use_virtual_board and
-                n_machine_time_steps is not None):
+                not self._use_virtual_board):
             algorithms.append("ChipIOBufExtractor")
 
         # add in the timing finalisation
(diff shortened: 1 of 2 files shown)


alan-stokes:
fair enough. can see why that would bugger it up. but why have both a class variable and it being handed down to _run? keep one and lose the other i would have thought

andrewgait:
> fair enough. can see why that would bugger it up. but why have both a class variable and it being handed down to _run? keep one and lose the other i would have thought

Will do this once @lplana has confirmed this fixes his issue...

andrewgait:
Fixed to use class variable rather than passing value down through functions.

Started testing on SpiNNakerManchester/SpiNNakerGraphFrontEnd#143 and SpiNNakerManchester/sPyNNaker8#402.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -279,6 +279,9 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
         # The lowest values auto pause resume may use as steps
         "_minimum_auto_time_steps",
 
+        #
```
alan-stokes, line 282: docs

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -492,6 +495,7 @@ def __init__(
         self._no_machine_time_steps = None
         self._minimum_auto_time_steps = self._config.getint(
                 "Buffers", "minimum_auto_time_steps")
+        self._run_until_complete = False
```
alan-stokes, line 498: id personally move that above the things where we start reading configs. but your welcome to ignore this one

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1931,7 +1934,7 @@ def _do_run(self, n_machine_time_steps, graph_changed, run_until_complete):
             reraise(*e_inf)
 
     def _create_execute_workflow(
-            self, n_machine_time_steps, graph_changed, run_until_complete):
+            self, n_machine_time_steps, graph_changed):
```
alan-stokes, line 1937: realign with def. as fits on 1 line now

alan-stokes:
after that. i consider this done

alan-stokes:
sorry bud. conflict.

andrewgait:
> after that. i consider this done

All done now I think... ?

PR #602, opened 2020-05-14
Proper reboot wait

alan-stokes:
start of the protection effort.


rowleya:
I'm guessing this is still in progress since it currently tells the user that they can configure the power wait time in the cfg, but uses a constant from SpiNNMan...  I suspect the cfg option is the right way to go here!

alan-stokes:
when i read that message. i dont get the view that it can change the wait time. but that they can avoid the wait by turning off the reboot.

alan-stokes:
but to cover your opinion. i thought this pr was done

rowleya:
Ah OK, I misread the message!  I do think it is a bit odd having a constant that is defined in SpiNNMan that is then only used in FEC...

alan-stokes:
there is likely other places in SPiNNMan where this constant can be used. I just went for the 1 location that was criticle for my work.

andrewgait:
This also appears to be having issues with one of the integration tests, though that could be a Travis problem...

rowleya:
It is worth noting that I think this wait only happens if the reset_machine_on_startup = True; I think this is False by default.

alan-stokes:
> Actually, I've changed my mind here. Do we really want to default wait for 30 seconds when running any script - that will add lots of unnecessary time to the integration tests, for a start... ? I'd suggest the cfg parameter should default to False and only be turned on where it's needed...

as pointed out by @rowleya if you dont have reset of startup. its not a problem. But if you do have reset of startup, it def is a thing which should reflect @lplana view that they need a 30 between power off and power on. so this would be the only sensible thing to do to enforce it.

on the intergration tests, just dont reset after startup. i persume the board per test is different so shouldnt need this?

andrewgait:
> It is worth noting that I think this wait only happens if the reset_machine_on_startup = True; I think this is False by default.

No, it's set to True by default at the moment (FEC cfg file line 172).

rowleya:
Oh, surprising.  I guess by default it tries to turn off a machine but doesn't have BMP access (on spalloc), so it doesn't get noticed.  In that case, maybe setting that parameter to False by default is the answer.

rowleya:
Alternatives include checking if this will have any effect and only pausing if it will

andrewgait:
> Alternatives include checking if this will have any effect and only pausing if it will

OK - will leave this for @alan-stokes to consider then.  At the moment every time I run a script on a 4-node board from this branch it waits for 30 seconds at the start...

alan-stokes:
> > It is worth noting that I think this wait only happens if the reset_machine_on_startup = True; I think this is False by default.
> 
> No, it's set to True by default at the moment (FEC cfg file line 172).

interesting. given the copy i have on the machine says its false. but once master is merged in. it gets shifted to true........... so thats a master change causing a bacl lash. fixing right now

alan-stokes:
> Alternatives include checking if this will have any effect and only pausing if it will

that is a bit harder to deal with. coz the spinnman power on and off dont tell you if they did anything......

alan-stokes:
tweaked spinnman a little, but now if you have a 4 chip board and thus no bmps but request a power cycle, it'll warn you to either turn off the request for power cycle, or use spalloc or give it bmp names in the cfg. It will not wait for 30s in this case. 

the default is not to power-cycle so the tests should be as fast as they were.

PR #604, opened 2020-05-22
Prevent a buffer overrun in the LPG when using packets with payloads

### LINKED ISSUE #593: Encountered an issue within the live_packet_gather.c file (flush_events seems to corrupt the circular_buffer 'with_payload_buffer')
Hi everybody,

currently I am trying to turn the Conways example from the [`SpiNNakerGraphFrontEnd`](https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/tree/master/spinnaker_graph_front_end/examples/Conways/partitioned_example_b_no_vis_buffer) into a version that streams the state of each cell for each time-step back to the host (you can find the code [here](https://github.com/jofas/master_thesis/tree/lpg_debug/learn_spinn)). For this I am using a `LivePacketGatherMachineVertex` instance. 

After debugging the program with @alan-stokes, we found out that the `flush_events` function somehow corrupts the `circular_buffer` '`with_payload_buffer`'. The following version of the `incoming_event_payload_callback` function:

```C
static void incoming_event_payload_callback(uint key, uint payload) {
    log_info("Received key %x, payload %x", key, payload);
    circular_buffer_print_buffer(with_payload_buffer);
    log_info("size of buffer: %d", circular_buffer_size(with_payload_buffer));

    if (circular_buffer_add(with_payload_buffer, key)) {

        log_info("buffer before adding: ");
        circular_buffer_print_buffer(with_payload_buffer);
        log_info("size of buffer: %d", circular_buffer_size(with_payload_buffer));

        circular_buffer_add(with_payload_buffer, payload);

        log_info("buffer after adding: ");
        circular_buffer_print_buffer(with_payload_buffer);
        log_info("size of buffer: %d", circular_buffer_size(with_payload_buffer));

        if (!processing_events) {
            processing_events = true;
            spin1_trigger_user_event(0, 0);
        }
    } else {
        provenance_data.number_of_overflows_with_payload++;

        log_info("overflow for key: %x with payload: %x", key, payload);
        circular_buffer_print_buffer(with_payload_buffer);
        log_info("size of buffer: %d", circular_buffer_size(with_payload_buffer));
        log_info("not full: %d", _circular_buffer_not_full(with_payload_buffer));
        log_info("not empty: %d", _circular_buffer_not_empty(with_payload_buffer));
    }
}
```

produces as log for a single call:

```
[INFO] (live_packet_gather.c: 319): Received key b, payload 0
[]
[INFO] (live_packet_gather.c: 321): size of buffer: 0
[INFO] (live_packet_gather.c: 342): overflow for key: b with payload: 0
[]
[INFO] (live_packet_gather.c: 344): size of buffer: 0
[INFO] (live_packet_gather.c: 345): not full: 0
[INFO] (live_packet_gather.c: 346): not empty: 0
```

This happens after ```flush_events``` is triggered the first time with a `buffer_index > 0`. Before the lgp works fine.
Once we commented out the content of the ```flush_events``` function, we didn't enter the `else` clause when we ran a simulation.
The `circular_buffer` somehow thinks it is full even though it is empty.

Also, using [this old version](https://raw.githubusercontent.com/SpiNNakerManchester/SpiNNFrontEndCommon/6bf9b6504b8d4bec62bae09bf745116c2f024a23/c_common/models/live_packet_gather/src/live_packet_gather.c) of the lpg worked once @alan-stokes made it use the newer APIs.

I hope this can help the team diagnose this and fix as quickly as possible.

Kind regards,
Jonas

dkfellows:
The overrun was because of a miscalculation of the number of words to clear after sending a message, yet it turns out that you don't need to clear the memory at all.

Also factor out all the actual writes (not of just the unaligned words) to helper inline functions to make the code easier to debug.

Fixes #593

### DIFF
--- a/c_common/models/live_packet_gather/src/live_packet_gather.c
+++ b/c_common/models/live_packet_gather/src/live_packet_gather.c
@@ -114,6 +114,9 @@ static struct lpg_config config;
 
 //! \brief Because WHY OH WHY would you use aligned memory? At least with this
 //! we don't get data aborts.
+//! \param[in] base: place to write to
+//! \param[in] index: location in base to write to
+//! \param[in] value: the value to write to the base.
 static inline void write_word(void *base, uint32_t index, uint32_t value) {
     uint16_t *ary = base;
     uint32_t idx = index * 2;
@@ -122,6 +125,9 @@ static inline void write_word(void *base, uint32_t index, uint32_t value) {
 }
 
 //! \brief Simple mirror of write_word() for true 16 bit values.
+//! \param[in] base: place to write to
+//! \param[in] index: location in base to write to
+//! \param[in] value: the value to write to the base.
 static inline void write_short(void *base, uint32_t index, uint32_t value) {
     uint16_t *ary = base;
     ary[index] = CLAMP16(value);
@@ -137,6 +143,7 @@ static inline uint8_t get_event_count(void) {
     return event_count;
 }
 
+//! \brief fills the SDP message and empties the buffer.
 static void flush_events(void) {
     // Send the event message only if there is data
     if ((buffer_index > 0) && (
@@ -169,6 +176,7 @@ static void flush_events(void) {
 }
 
 //! \brief function to store provenance data elements into SDRAM
+//! \param[in] provenance_region_address: address for storing provenance
 static void record_provenance_data(address_t provenance_region_address) {
     struct provenance_data_struct *sdram = (void *) provenance_region_address;
     // Copy provenance data into SDRAM region
@@ -176,6 +184,9 @@ static void record_provenance_data(address_t provenance_region_address) {
 }
 
 // Callbacks
+//! \brief timer that clears packets stored and stops if required.
+//! \param[in] unused0: api limit.
+//! \param[in] unused1: api limit.
 static void timer_callback(uint unused0, uint unused1) {
     use(unused0);
     use(unused1);
@@ -202,13 +213,16 @@ static void timer_callback(uint unused0, uint unused1) {
     }
 }
 
+//! \brief checks if there's enough stored events to push to a SDP message
+//! and if so, fills and sends the SDP message.
 static inline void flush_events_if_full(void) {
     if ((get_event_count() + 1) * event_size > BUFFER_CAPACITY) {
         flush_events();
     }
 }
 
-// processes an incoming multicast packet without payload
+//! \brief processes an incoming multicast packet without payload
+//! \param[in] key: multicast key.
 static void process_incoming_event(uint key) {
     log_debug("Processing key %x", key);
 
@@ -237,7 +251,9 @@ static void process_incoming_event(uint key) {
     }
 }
 
-// processes an incoming multicast packet with payload
+//! \brief processes an incoming multicast packet with payload
+//! \param[in] key: multicast key.
+//! \param[in] payload: multicast payload.
 static void process_incoming_event_payload(uint key, uint payload) {
     log_debug("Processing key %x, payload %x", key, payload);
 
@@ -267,6 +283,10 @@ static void process_incoming_event_payload(uint key, uint payload) {
     }
 }
 
+//! \brief user interrupt for pulling out of the buffer and filling up a SDP
+//! message. Will sned SDP if more than 1 SDP message needed.
+//! \param[in] unused0: api limit.
+//! \param[in] unused1: api limit.
 static void incoming_event_process_callback(uint unused0, uint unused1) {
     use(unused0);
     use(unused1);
@@ -289,6 +309,9 @@ static void incoming_event_process_callback(uint unused0, uint unused1) {
     } while (processing_events);
 }
 
+//! \brief callback for a multicast packet without a payload.
+//! \param[in] key: the multicast key.
+//! \param[in] unused: api limit.
 static void incoming_event_callback(uint key, uint unused) {
     use(unused);
     log_debug("Received key %x", key);
@@ -303,6 +326,9 @@ static void incoming_event_callback(uint key, uint unused) {
     }
 }
 
+//! \brief callback for a multicast packet with a payload.
+//! \param[in] key: the multicast key.
+//! \param[in] payload: the payload of the multicast packet.
 static void incoming_event_payload_callback(uint key, uint payload) {
     log_debug("Received key %x, payload %x", key, payload);
 
@@ -317,6 +343,8 @@ static void incoming_event_payload_callback(uint key, uint payload) {
     }
 }
 
+//! \brief init method.
+//! \param[in] sdram_config: sdram location for the lpg parameters.
 static void read_parameters(struct lpg_config *sdram_config) {
     // Faster to copy by field than to use spin1_memcpy()!
 
@@ -356,6 +384,9 @@ static void read_parameters(struct lpg_config *sdram_config) {
     log_info("packets_per_timestamp: %d", config.packets_per_timestamp);
 }
 
+//! \brief top level init method.
+//! \param[in] timer_period pointer to int where timer period will be stored.
+//! \return bool where True was successful init and  false otherwise.
 static bool initialize(uint32_t *timer_period) {
     // Get the address this core's DTCM data starts at from SRAM
     data_specification_metadata_t *ds_regions =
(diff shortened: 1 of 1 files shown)


c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -113,52 +114,54 @@ static struct lpg_config config;
...
-                    write_unaligned(sdp_msg_aer_payload_prefix, 0, time);
-                }
-            }
-
-            spin1_send_sdp_msg(&g_event_message, 1);
-            packets_sent++;
         }
 
-        // reset packet content
-        uint16_t words_to_clear = (buffer_index * event_size) >> 2;
```
dkfellows, line 158: This line was the actual place of the bug; when there are payloads, `buffer_index` is twice as large as when there are no payloads, so `words_to_clear` ends up twice as large as it should be. With enough packets, we end up overwriting memory outside the buffer…

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -184,6 +187,9 @@ static void timer_callback(uint unused0, uint unused1) {
     time++;
     log_debug("Timer tick %u", time);
 
+    // Reset the count of packets sent in the current timestep
+    packets_sent = 0;
```
dkfellows, line 191: This fixes a bug where setting `config.packets_per_timestamp` to non-zero would result in that being the maximum total number of packets ever sent, which was a dumb and non-useful feature! The problem was that the counter (`packets_sent`) was never being reset at the start of the next timestep.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -450,64 +447,45 @@ static bool configure_sdp_msg(void) {
     temp_header |= config.payload_timestamp << PAYLOAD_IS_TIMESTAMP;
     temp_header |= config.packet_type << PACKET_TYPE;
 
-    header_len = 2;
-
     // pointers for AER packet header, prefix and data
-    void *temp_ptr;
```
dkfellows, line 456: We don't need this. We can just use `sdp_msg_aer_data` and bump it along as required.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -450,64 +447,45 @@ static bool configure_sdp_msg(void) {
...
-        sdp_msg_aer_key_prefix = NULL;
-        temp_ptr = &sdp_msg_aer_header[1];
+        // pointer to key prefix, so data is one half-word further ahead
+        write_short(sdp_msg_aer_header, 1, config.prefix);
+        sdp_msg_aer_data++;
     }
 
     if (config.payload_apply_prefix) {
-        sdp_msg_aer_payload_prefix = temp_ptr;
-        uint16_t *a = sdp_msg_aer_payload_prefix;
```
dkfellows, line 470: I cannot remember what value `a` brought to the show. Purged as unnecessary.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -450,64 +447,45 @@ static bool configure_sdp_msg(void) {
...
-            log_debug("32 bit - temp_ptr: %08x", temp_ptr);
         }
     } else {
         sdp_msg_aer_payload_prefix = NULL;
     }
 
-    // pointer to write data
-    sdp_msg_aer_data = temp_ptr;
+    // compute header length
+    header_len = (sdp_msg_aer_data - sdp_msg_aer_header) * sizeof(uint16_t);
```
dkfellows, line 482: We know that the data starts straight after the header, and we've kept track of where the header starts and the data starts, so we might as well just use those pointers to compute the header length directly rather than trying to calculate them as we go along.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -113,52 +114,54 @@ static struct lpg_config config;
...
-        uint8_t event_count;
-
-        if ((config.packets_per_timestamp == 0)
-                || (packets_sent < config.packets_per_timestamp)) {
-            // Get the event count depending on if there is a payload or not
-            if (HAVE_PAYLOAD(config.packet_type)) {
-                event_count = buffer_index >> 1;
+    if ((buffer_index > 0) && (
+            (config.packets_per_timestamp == 0) ||
+            (packets_sent < config.packets_per_timestamp))) {
```
dkfellows, line 144: It turns out that combining the two `if`s is now trivial.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -113,52 +114,54 @@ static struct lpg_config config;
...
 }
 
+//! \brief Simple mirror of write_word() for true 16 bit values.
+static inline void write_short(void *base, uint32_t index, uint32_t value) {
+    uint16_t *ary = base;
+    ary[index] = CLAMP16(value);
+}
+
+//! \brief Get how many events there are waiting to be sent
+static inline uint8_t get_event_count(void) {
```
dkfellows, line 131: We put this calculation in one place. We do it correctly once so that we don't have the possibility of changing it in one place but not the other.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -113,52 +114,54 @@ static struct lpg_config config;
...
-        void *base, uint32_t index, uint32_t value) {
+static inline void write_word(void *base, uint32_t index, uint32_t value) {
     uint16_t *ary = base;
     uint32_t idx = index * 2;
     ary[idx++] = CLAMP16(value);
     ary[idx] = CLAMP16(value >> 16);
 }
 
+//! \brief Simple mirror of write_word() for true 16 bit values.
+static inline void write_short(void *base, uint32_t index, uint32_t value) {
```
dkfellows, line 125: Not an important function, except that it makes the code dealing with 16-bit values look like the code dealing with 32-bit values, which makes reading what is going on _much_ easier.

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -81,10 +82,10 @@ enum packet_types {
 // Globals
 static sdp_msg_t g_event_message;
 static uint16_t *sdp_msg_aer_header;
-static uint16_t *sdp_msg_aer_key_prefix;
```
dkfellows, line 84: Turns out we don't need `sdp_msg_aer_key_prefix` after the configuration phase.

dkfellows:
There we go, that's my annotated thoughts on the changes. Everything else should probably follow from those.

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: having looked at it (noted its disappeared in the new version) im not sure how the heck that ever worked for anything OTHER than things with event_size = 4. if i understand correctly. buffer_index is basically how many uint16's are in the circular buffer (which is likely the same as buffer->real_size / 4) so unless the mul and the shift cancel each other out. i can def see how that will misalign quickly. 

I'm going to presume that we got lucky for so long.

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: guessing thats a legacy debugging tool

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: yep

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: yea

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: yep

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: definately

c_common/models/live_packet_gather/src/live_packet_gather.c
alan-stokes: yep

alan-stokes:
par from missing docs on functions (which im putting in for ya now) im happy with this

dkfellows:
I was mostly leaving the docs alone; that's a different branch.

PR #605, opened 2020-05-22
Convert ybug to Python

dkfellows:
This is a conversion of the `ybug` tool into Python (to create a tool called `pybug`, short for “Python `ybug`”).

[EDIT]: Changed to `spybug`. No other tool is called that.


dkfellows:
What this is _not_ is a conversion of the code to use the rest of our toolchain. I've deliberately not done that.

tools/cmd.py
```diff
@@ -0,0 +1,510 @@
...
+    LINK_WRITE = 18
+
+    LED = 25
+    IPTAG = 26
+
+    # Tubotron message
+    TUBE = 64
+
+
+class SCAMP_CMD(CMD):
```
Christian-B, line 57: Computer says NO!
TypeError: Cannot extend enumerations
https://docs.python.org/3/library/enum.html#restricted-subclassing-of-enumerations

tools/cmd.py
```diff
@@ -0,0 +1,510 @@
...
+    ------------------------dddddddd   <<  0
+    """
+    return (
+        (a & 0xFF) << 24) | ((b & 0xFF) << 16) | ((c & 0xFF) << 8) | (d & 0xFF)
+
+
+# -----------------------------------------------------------------------------
+
+
+class CMD(IntEnum):
```
Christian-B, line 40: Not a nice name if we plan on using https://docs.python.org/3.8/library/cmd.html

tools/pybug.py
```diff
@@ -0,0 +1,1544 @@
...
+from tools.exn import BadArgs, SpinnException, SpinnTooManyRetriesException
+from tools.cli import CLI
+from tools.boot import boot
+from tools.sv import Struct
+from tools.cmd import SCAMPCmd, BMPCmd, SCAMP_CMD
+from tools.util import (
+    read_file, hex_dump, parse_cores, parse_region, parse_apps, parse_bits)
+
+# This code ought to be rewritten to use the Cmd package:
+# https://docs.python.org/3.8/library/cmd.html
```
Christian-B, line 36: Why not do it now? 

And if the answer is because it is messy with python2 then I vote that this new feature is not supported in python 2

tools/cli.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def arg_i(self, n):
+        return int(self._args[n], base=0)
+
+    def arg_x(self, n):
+        return int(self._args[n], base=16)
+
+    # Default commands of a CLI object; always present!
+
+    def _Pause(self):
```
Christian-B, line 128: Why underscore Captial name?  Same elsewhere

tools/pybug.py
```diff
@@ -0,0 +1,1544 @@
...
+def init_readline():
+    if not _readline:
+        return None
+
+    import readline
+    readline.set_completer(Completer(readline))
+    return readline
+
+
+def main():
```
Christian-B, line 1533: Needs minimal documentation on how to run.

This can/or should be limited to how to reach the internal help.

Equivellent of ybug -help

tools/pybug.py
```diff
@@ -0,0 +1,1544 @@
...
+        if word and len(self._stored) == 1:
+            word += " "
+        return word
+
+
+def init_readline():
+    if not _readline:
+        return None
+
+    import readline
```
Christian-B, line 1528: Why an import inside the code?

May be a code reeason but needs a comment then!

tools/pybug.py
```diff
@@ -0,0 +1,1544 @@
...
+
+def open_targets():
+    global spin, sv, bmp
+    spin = SCAMPCmd(target=spinn_target, port=spin_port, debug=debug)
+    sv = Struct(scp=spin)
+    if bmp_target is not None:
+        bmp = BMPCmd(target=bmp_target, port=bmp_port, debug=debug)
+
+
+class Completer(object):
```
Christian-B, line 1505: Inner class?

1. Does it need to be an inner Class
2. Use a underscore name.

tools/cmd.py
dkfellows: Fixed

tools/cmd.py
dkfellows: Almost all of the names match ones from the ybug code. Which isn't to say that they shouldn't change.

tools/cli.py
dkfellows: Capital to match ybug, `_` because it's internal to the class (it was added later).

tools/pybug.py
dkfellows: In this case, it's matching the squirrels in ybug.

dkfellows: Also, care needs to be taken because the initialisation of the readline library should only occur if the command line options don't tell it not to.

tools/pybug.py
dkfellows: > Use a underscore name.

This.

tools/pybug.py
rowleya: In ybug, we import readline if we can (as that makes it easier to use).  This is because we don't "install" ybug as such, so it never gets a chance to install dependencies, and readline isn't normally installed with perl (i.e. by default).  In python this isn't necessary, but I understand that this is a proper port of the code, so understandable.  This can be removed in a future refactor, and just add a dependency on readline (assuming that makes it possible to go to past commands in Python).

tools/cli.py
dkfellows: Changed `_Pause` to `_cmd_pause` (and analogous other methods).

tools/pybug.py
dkfellows: Added

tools/pybug.py
dkfellows: The important question here is whether we need to support disabling readline in the first place? If not, then we can import it always (merely importing it also turns it on, which is why this is done the way it is right now). Then, there's separately the fun of trying to make a better `_Completer` class; what's there right now is a pale shadow of what the Perl code does because there's no fallback filename completion operation in the Python `readline` library.

But that's something we can do or not do as we see fit. Filename completion isn't very useful for most of the things we use ybug for these days.

tools/cmd.py
dkfellows: `CMD` &rarr; `SCP_CMD`

tools/pybug.py
rowleya: No, it doesn't need to support being disabled here as python has readline support by default as I understand it.  I think I have used filename completion before, but agree this isn't a priority.

tools/pybug.py
dkfellows: I think I'd describe that as a longer-term plan. It's quite a large change as it works in a different way.

I found that package when looking up how to make readline (itself not one of the better documented Python packages) work.

tools/pybug.py
dkfellows: Removed the optionality of the readline support.

tools/pybug.py
dkfellows: Since this is a disguised call for doing a lot of work, I _reject_ your comment as relevant for this PR.

setup.py
```diff
@@ -22,24 +22,26 @@
 assert __version__
 
 # Build a list of all project modules, as well as supplementary files
-main_package = "spinn_front_end_common"
+main_packages = ("spinn_front_end_common", "tools")
```
Christian-B, line 25: I would change the main package from "tools" to "spybug" and from tools import   is too generic

Christian-B (changes requested):
I would not use "tools" as a main package name.

That makes imports hard.

Use spybug instead

setup.py
alan-stokes: spy implies only pynn.......

dkfellows:
> I would not use "tools" as a main package name.

I agree. It's now in `spinn_front_end_common.tools.spybug`

dkfellows:
Not important to me any more.

PR #609, opened 2020-06-10
constants moved to Pacman

Christian-B:
Avoid duplication of the constants moved in https://github.com/SpiNNakerManchester/PACMAN/pull/297   

Other classes can still import SARK_PER_MALLOC_SDRAM_USAGE from here as it is still there as it is needed for other constants anyway

Can be merged after but not before https://github.com/SpiNNakerManchester/PACMAN/pull/297


spinn_front_end_common/utility_models/multi_cast_command.py
```diff
@@ -1,127 +1,127 @@
...
-        """
-        return self._payload is not None
-
-    def __repr__(self):
-        return \
-            "MultiCastCommand(time={}, key={}, payload={},"\
-            " time_between_repeat={}, repeats={})".format(
-                self._time, self._key, self._payload,
-                self._delay_between_repeats, self._repeat)
+# Copyright (c) 2017-2019 The University of Manchester
```
dkfellows, line 1: Why are there any changes to this file at all? (Has something decided to change line endings on us?)

unittests/interface/interface_functions/test_output/core1.json
```diff
@@ -0,0 +1,5 @@
+{
```
dkfellows, line 1: Why do we have `test_output` in here?

alan-stokes:
conflicts and open issues

Christian-B:
old an unmerged so assuming no longer needed

PR #610, opened 2020-06-11
Better logging of what the reinjector is doing

dkfellows:
We want the Data in/out to be quiet, but reinjector control is important and much less common.

This PR just adds some more prints to IOBUF.


dkfellows:
Nothing important enough here to be worth keeping this branch going for

PR #613, opened 2020-06-11
Unaligned access trickery

dkfellows:
Live Packet Gatherer: A neater way to write the `write_word()` code. This results in two half-word store instructions being generated (which is the ideal case given that a full word store results in a bus error) while still being recognisably an actual write of a word.


alan-stokes:
can this have a test suite associated with it. when anything RIPTMCS changes, i never trust it without full tests having gone over it

PR #616, opened 2020-06-17
Graph mapper2

Christian-B:
Refactor EdgeToNKeysMapper so it can handle graphs with some but not all Machine Vertexes have application vertextes.


Christian-B:
Discussed EdgeToNKeysMapper with @alan-stokes and after agreeding pushed straight to Graph Mapper branch.
All other files should not be here and github is confused so closing this PR.

See:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/commit/ff7a4664c6ea96feea7056fc771fd3ead5190879

Issue #623, opened 2020-06-29
ReverseIPTagMulticastSourceMachineVertex has different  meanings for send_buffer_time is None


Christian-B:
The class ReverseIPTagMulticastSourceMachineVertex has different  meanings for send_buffer_time is None
And this effects the sdram cost for recording

in method _recording_sdram_per_timestep(...
the assumption is send_buffer_times is None means  recording live data

in the init and empty  send_buffer_times is converted to None.  
However an empty  send_buffer_times param does not indicate live data.

Caught in spy8 integration test  TestSpikeSourceArray.test_recording_with_empty_lists
IF ReverseIpTagMultiCastSource.create_machine_vertex(....
has an extra
        if resources_required:
            assert (resources_required == machine_vertex.resources_required)

rowleya:
I assume you mean what happens if someone creates a SpikeSourceArray with no spikes and then asks to record it?  I can see why that would go wrong...

alan-stokes:
well...........what happens when you say use it for spike injection?

rowleya:
From memory alone I think you can record live spike input; it reserves "some space" for that and then fails to record if you go over that space.  I believe that this is user controlled, but again only from memory.

rowleya:
The issue here though is that an empty SSA with recording should require very little SDRAM, but it probably reserves enough for live spikes in that case as it doesn't see the empty SSA as being predetermined and fixed.

alan-stokes:
if it rserved enough for live spikes. it should still agree with the app vertex resources. the issue is that it doesn't. 

thus ReverseIPTagMulticastSourceMachineVertex's resources_required
and 
ReverseIpTagMultiCastSource's get_resources_used_by_atoms

dont match results, Which is bad bad

alan-stokes:
and funnily. ReverseIPTagMulticastSourceMachineVertex's  get_sdram_usage and corrasponding dtcm and cpu calls are static method. So why the app vertex makes its own logic...............

rowleya:
Yes odd - the app vertices are supposed to use the machine vertex in that case I believe

Christian-B:
To be clear the issue is caused not by the sdram cost function which as @rowleya points out is the same function but by how send_buffer_times is handled. IE what is passed into that method

In the init of ReverseIPTagMulticastSourceMachineVertex there is code:
        if n_buffer_times == 0:
            self._send_buffer_times = None

[...]

PR #625, opened 2020-07-07
Deal with case where a vertex_slice has an array of empty arrays

andrewgait:
Fixes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/623

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/416

### DIFF
--- a/spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
+++ b/spinn_front_end_common/utility_models/reverse_ip_tag_multi_cast_source.py
@@ -179,6 +179,17 @@ class ReverseIpTagMultiCastSource(
             if hasattr(send_buffer_times[0], "__len__"):
                 send_buffer_times = send_buffer_times[
                     vertex_slice.lo_atom:vertex_slice.hi_atom + 1]
+                # Check the buffer times on the slice are not empty
+                n_buffer_times = 0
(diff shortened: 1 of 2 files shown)


Christian-B:
I think this fixes it the wrong way around.

Now both the Application Vertex and the Machine vertex trigger the block 
which bases sdram on the receive_rate

This is incorrect as there in no live data.

It is incorrect to assume there will be on the one core whose splice has no values in the send_buffer_times.

So while the resources require now match they are now both wrong

Christian-B:
I have added a log warning to track that there is a weirdness going on.

andrewgait:
> I think this fixes it the wrong way around.
> 
> Now both the Application Vertex and the Machine vertex trigger the block
> which bases sdram on the receive_rate
> 
> This is incorrect as there in no live data.
> 
> It is incorrect to assume there will be live data on the one core whose splice has no values in the send_buffer_times.
> 
> So while the resources require now match they are now both wrong

I think you're right, but it's not necessarily straightforward to fix it the other way - it certainly involves more changes, maybe even to the buffer manager.  As you say maybe it's better to fix it this way but have a log warning.

alan-stokes:
keeping branch for any reason?

PR #628, opened 2020-07-09
Untangle

Christian-B:
Alternative to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/626 which uses one progress bar for both prepare and check for success.

Do not merge both they clash.

Depends on:
https://github.com/SpiNNakerManchester/SpiNNMan/pull/212
Used by:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/838

Includes 
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/commit/e89c0da675cf948d9d1dceaf4f85beb137d2dd0e
Which fixes: https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/627

PR #631, opened 2020-07-14
Pair untangle

Christian-B:
Builds on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/628

Combines 4 compressors into shared code
bit_feild unordered
bit_feild pair
simple unordered
simple pair

Depends on https://github.com/SpiNNakerManchester/SpiNNMan/pull/212
used by https://github.com/SpiNNakerManchester/sPyNNaker/pull/841
tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/420 (please merge too)


alan-stokes:
I honestly think this is way overkill. I dont grasp why we need two apis for routing_table.h and why we cant just have a bool flag from init decide which top control logic to use. thus making 2 binaries instead of 4. This looks well too complex. I propose asking @rowleya for a second view. 

otherwise the python looks fine.

Christian-B:
The simple method has a single routing table on which it operates in place.

The bitfeild for each attempt has to make a copy of the routing table which to make sure the multiple copies fit are split into multiple tables,

Why oh why make a more complicated copy just to use to keep avoid having two solutions.

alan-stokes:
to avoid 2 solutions. obviously. the speed up shouldn't be that significant. But like i said, ill bow to a 3rd opinion.

rowleya:
If you can point me at the bit you are worried about in the 42 changed files, I can have a look...

alan-stokes:
its all in the c code to do with the compressors and the use of the routing_table.h/c being two differnet impls making 4 binaries. basically i think theres a cleaner way to do without all the gubbins. but as @Christian-B states, my way would be a bit slower. but id argue cleaner to read.

rowleya:
OK, but there are only two routing_table.h files, and one imports from the other right?  Or am I reading this wrong?
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/bit_field_common/routing_tables.h
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/common/routing_table.h

rowleya:
I have had a look at the files with c_main:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/simple/simple_compressor.c
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/bit_field_compressor.c

There seems to be little commonality between those, so I can't see them being merged in a simple way.

The other thing that makes two binaries is the use of USE_PAIR.  As this brings in an entirely different compression algorithm, I can also see why that would be a different binary.  If we ever add another compression algorithm, a joint file with a config option would be starting to get quite big...

alan-stokes:
> OK, but there are only two routing_table.h files, and one imports from the other right? Or am I reading this wrong?
> https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/bit_field_common/routing_tables.h
> https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/3eaa72661a8841c5a3d23c5cd5527438961abfd5/c_common/models/compressors/src/common/routing_table.h

thats my point. you can instead of haivng 2 impls. you could easily have 1 impl and have the host load the data in the common format then.

alan-stokes:
the reason you dont see much commonality is coz as it CURRENTLY stands, they dont. But a wee tweek on the host and you could get both to work with the same original routing_table.h used by the compressor and sorter. And when you do that, you dont even need to have 4 binairies, you can have 2. 

"The other thing that makes two binaries is the use of USE_PAIR. As this brings in an entirely different compression algorithm, I can also see why that would be a different binary. If we ever add another compression algorithm, a joint file with a config option would be starting to get quite big..."

im not arguing for 1 binary. im arguing for 2. a pair and a mundy. where each one handles both use as a compressor in the sorter compressor tie in AND as a stand alone. the FACT that you have them working in the sorter compressor mode means that ITCM isnt a issue. as All your saying is a flag which BYPASSES all the gubbins the sorter compressor api needs, and just instead loads the sdram directly and bobs ya uncle  run as normal.

rowleya:
I'm not sure what you gain from that.  You would have a single C file that has two purposes I think, as one is run by itself, and one runs as a service to the searcher (if I understand how it works).  Yes, you would save some number of Kb (<32kB by design) in the final distribution, but I don't see that as being critical given what we already do in the neural code.

alan-stokes:
well if your happy with these ............. complexity. then i'll approve it. i still hold the cake in case i get to throw it at you in future.

rowleya:
Yes, I am happy with the reduced complexity of having two different c_mains for two different use cases.

PR #632, opened 2020-07-15
Determism

alan-stokes:
work to make all networks more repeatable.


rowleya:
Question (so I know what I am looking at): is it ever likely that a user of the TDMA in the C code will ever call the function in two different parts of the code (and if so, do you have an example of this happening)?

alan-stokes:
so I'm going to interpret "example" and "ever" as "can you name a use case for multiple calls". Then i propose calcium channels, dopamine channels. neurons spiking for those with different keys and thus TDMA slots.

rowleya:
I think that they are cases for multiple reception rather than multiple transmission.  The reason I asked was in an attempt to determine whether this code would be better inlined (i.e. in a header so it can be inlined).  I will have a look in any case.

alan-stokes:
oooooohhhhhh then i point out possion and delay

alan-stokes:
they call it twice. 1 for no payload, and 2 for payload. so if you inline it, you'll double the code there.

rowleya:
OK, no problem; this is why I wanted to check :)

alan-stokes:
sorry that i didn't get what you were leading to the first time. who knows, maybe you'll come up with some black magic to remove those 2 calls to 1. id be amazed, but it might be possible.

spinn_front_end_common/abstract_models/abstract_requires_tdma.py
```diff
@@ -0,0 +1,94 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from six import add_metaclass
+from spinn_utilities.abstract_base import (
+    AbstractBase, abstractmethod, abstractproperty)
+
+
+@add_metaclass(AbstractBase)
+class AbstractRequiresTDMA(object):
```
dkfellows, line 22: Why do you have this interface? It only has one direct implementation, and there doesn't appear to be any prospect of another one ever being created. Roll it into `RequiresTDMA` and don't proliferate interfaces that don't add value.

spinn_front_end_common/abstract_models/impl/requires_tdma.py
```diff
@@ -0,0 +1,108 @@
...
+from spinn_front_end_common.utilities import constants
+from spinn_front_end_common.utilities.utility_objs import ProvenanceDataItem
+from spinn_utilities.overrides import overrides
+
+
+# The number of clock cycles per nanosecond
+_CLOCKS_PER_NS = 200
+
+
+class RequiresTDMA(AbstractRequiresTDMA):
```
dkfellows, line 29: This class has a slots problem. It needs its concrete subclasses (`AbstractPopulationVertex`, `SpikeSourcePoissonVertex`, and `DelayExtensionVertex`) to provide a very specific set of fields… and yet needs to have empty slots because of multiple inheritance. Something's very much not right. Perhaps making a `TDMAAwareApplicationVertex` (since all known uses are application vertices) instead of having this be totally separate would be a better approach?

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
```diff
@@ -0,0 +1,283 @@
...
+from spinn_front_end_common.utilities import helpful_functions, \
+    globals_variables
+from spinn_front_end_common.utilities.exceptions import ConfigurationException
+from spinn_utilities.logger_utils import warn_once
+
+logger = logging.getLogger(__name__)
+
+
+class LocalTDMABuilder(object):
+    """ Builds a localised TDMA which allows a number of machine vertices
```
dkfellows, line 30: Have you checked how Sphinx renders this comment?

spinn_front_end_common/utilities/helpful_functions.py
```diff
@@ -171,6 +171,22 @@ def read_config_int(config, section, item):
     return int(value)
 
 
+def read_config_float(config, section, item):
```
dkfellows, line 174: I like this.

c_common/front_end_common_lib/include/tdma_processing.h
```diff
@@ -0,0 +1,63 @@
...
+    uint32_t min_expected_time;
+    //! The time between sending
+    uint32_t time_between_sends;
+} tdma_parameters;
+
+//! \brief hands back the number of times the TDMA was behind
+uint32_t tdma_processing_times_behind(void);
+
+//! \brief init for the tdma processing
+//! \param[in/out] address: pointer to the SDRAM address where this data is
```
dkfellows, line 45: It's `\param[in,out]`; alas, Doxygen is picky about this.

c_common/front_end_common_lib/include/tdma_processing.h
```diff
@@ -0,0 +1,63 @@
...
+//!                         stored, updated after being read
+//! \return: bool saying success or fail
```
dkfellows, line 47: Phrasing: `\return whether we succeeded` (unlike with Python, the type is obvious)

c_common/front_end_common_lib/include/tdma_processing.h
```diff
@@ -0,0 +1,63 @@
...
+typedef struct tdma_parameters {
+    //! The time at which the first message can be sent
+    uint32_t initial_expected_time;
+    //! The time at which the last message must be sent by
+    uint32_t min_expected_time;
+    //! The time between sending
+    uint32_t time_between_sends;
+} tdma_parameters;
+
+//! \brief hands back the number of times the TDMA was behind
```
dkfellows, line 41: Needs a two-line documentation comment, perhaps this:
```c
//! \brief Get the number of times that the TDMA was behind
//! \return the number of times the TDMA lagged
```

dkfellows (changes requested):
There's too much inheritance trouble going on in the Python code. Please put a stop to it.

Also a bunch of minor doc issues that won't take more than a few moments to fix. 😉

dkfellows:
Overall idea seems like a decent plan. Details need work.

spinn_front_end_common/abstract_models/impl/requires_tdma.py
alan-stokes: perfectly happy to do this if thats how we want to go.

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
alan-stokes: have i checked how anything is rendered in sphinx. how would i go about checking?

c_common/front_end_common_lib/include/tdma_processing.h
alan-stokes: fixed

c_common/front_end_common_lib/include/tdma_processing.h
alan-stokes: done

c_common/front_end_common_lib/include/tdma_processing.h
alan-stokes: done

spinn_front_end_common/abstract_models/abstract_requires_tdma.py
alan-stokes: so where ill agree given the removal via new type of app vertex. i feel ...... inclined to point out that there is indeed a second TDMA implementation floating around our tool chain. It just doesn't use any API's and assumes all vertices will use its generated agenda. it exists in fecs->interface_functions->tdma_agenda_builder. but that one ensures 1 packet on the ENTIRE network of interconnecting vertices. but noone has used it for an age

spinn_front_end_common/abstract_models/impl/requires_tdma.py
alan-stokes: done

spinn_front_end_common/abstract_models/abstract_requires_tdma.py
alan-stokes: done

dkfellows:
XREFs:

* SpiNNakerManchester/sPyNNaker#842
* SpiNNakerManchester/sPyNNaker8#421
* SpiNNakerManchester/PyNN8Examples#39

alan-stokes:
> XREFs:
> 
> * [SpiNNakerManchester/sPyNNaker#842](https://github.com/SpiNNakerManchester/sPyNNaker/pull/842)
> * [SpiNNakerManchester/sPyNNaker8#421](https://github.com/SpiNNakerManchester/sPyNNaker8/pull/421)
> * [SpiNNakerManchester/PyNN8Examples#39](https://github.com/SpiNNakerManchester/PyNN8Examples/pull/39)

you can find a full mapping in the determism project here:
https://github.com/orgs/SpiNNakerManchester/projects/35

rowleya (approved):
This is fine if the docs are OK as suggested by Donal

alan-stokes:
> This is fine if the docs are OK as suggested by Donal

doanl approved via skype. so no holding it back

PR #633, opened 2020-07-16
router diagnositcs misleading

alan-stokes:
this clears them the moment before app runner sends the signal to run. anytime before this is counter productive as otehr loading algorithms could end up producing packet drops.


alan-stokes:
testing via: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/422

PR #642, opened 2020-07-28
compressor cleanup

Christian-B:
1. Stops run is there are no routing tables. Thanks @andrewgait 
2. Only writes iobuf is cfg write_compressor_iobuf is overwritten to true
    - mode debug will also set that to true
    - All compressors use the same cfg setting 
3. If compressor fails will extract iobuf

Must be done at same time as https://github.com/SpiNNakerManchester/sPyNNaker/pull/851

tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/428


Christian-B:
ChipPowerMonitor chance creeped in sorry, but is not wrong as Machine vertex has same functionality

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
```diff
@@ -36,7 +36,7 @@
 def mundy_on_chip_router_compression(
         routing_tables, transceiver, machine, app_id,
         system_provenance_folder, compress_only_when_needed=True,
-        compress_as_much_as_possible=False):
+        compress_as_much_as_possible=False, write_compressor_iobuf=False):
```
alan-stokes, line 39: this isnt really sensible. its never called directly, and the 2 places that do call it take the write_compressor_iobuf as a required parameter. so having a default value here makes no sense.

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
```diff
@@ -56,22 +56,23 @@ def mundy_on_chip_router_compression(
...
+        write_compressor_iobuf)
     compression._compress_only_when_needed = compress_only_when_needed
     compression.compress(register=0)
 
 
 def pair_compression(
         routing_tables, transceiver, executable_finder,
         machine, app_id, provenance_file_path,
-        compress_as_much_as_possible=True):
+        compress_as_much_as_possible=True, write_compressor_iobuf=False):
```
alan-stokes, line 75: same problem here

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
```diff
@@ -91,21 +92,23 @@ def pair_compression(
...
+        "Running pair routing table compression on chip",
+        write_compressor_iobuf)
     compression.compress(register=1)
 
 
 def unordered_compression(
         routing_tables, transceiver, executable_finder,
         machine, app_id, provenance_file_path,
-        compress_as_much_as_possible=True):
+        compress_as_much_as_possible=True, write_compressor_iobuf=False):
```
alan-stokes, line 111: and here

alan-stokes (changes requested):

alan-stokes:
apart from the default param when it cant be defaulted, i actually like this pr.

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
alan-stokes: fixed it for ya

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
alan-stokes: fixed

spinn_front_end_common/interface/interface_functions/on_chip_router_table_compression/compression.py
alan-stokes: cleaned up for ya

PR #647, opened 2020-08-18
 Create on the fly data databse for recorded data

Christian-B:
First version of a create on the fly database to hold recorded data.

Suggestion is to run unittests/interface/buffer_management/test_recorded_database.py
and examine the resulting test,sqlite3 file

Next steps
1. Add a tool to create neo segments from the database
2. At the end of run and reset ssave the data to the database instead of caching it in ram
3 On get data get it from the database

Future work.
1. Add more datatypes


Christian-B:
The current view only creates some viewes if the data return is different in the view.]
Or should we create all views all the time?

Christian-B:
@dkfellows questioned why the local id is NULl in some cases.
The reason is that for some variables like spikes there is no benefit need to split the data into one table per core.

It would be possible but then we would have to ask the data provider to give the lowest possible neuron id as just getting it from the data is not possible as different neurons may spike in different runs.

dkfellows:
At the moment, I'm trying to comprehend the details and the use cases. My observation is that there's lots of tables and no indices. That's only ever acceptable when the sole use cases are append and reading the whole lot back via linear scan. I'm wondering whether having fewer basic tables with suitable indices would be better, and then constructing views on top of that to get the recordings-as-simple-tables vibe. (I know how to configure triggers so that the views are write-through, if that's required, but it's possible that it isn't.)

dkfellows:
Also wondering whether the general recording data type should be an enum.

dkfellows:
Something's going wonky. In the test DB, table `pop1_spikes_ALL_raw` has the unlikely column name `atom idtimestamp` which seems rather unlikely.

dkfellows:
More seriously, I don't actually understand the data that is in that table. (The other tables and views in the DB generated by the tests look OK to me. While they might end up rearranged, they don't look _wrong._)

spinn_front_end_common/interface/buffer_management/recorded/recorded_database.py
```diff
@@ -0,0 +1,597 @@
...
+        table_name = self._table_name(source_name, variable_name, RAW)
+        cursor.execute(
+            """
+            INSERT INTO local_metadata(
+                source_name, variable_name, table_name, first_neuron_id,
+                data_type)
+            VALUES(?,?,?,?,?)
+            """, (source_name, variable_name, table_name, None, EXISTS))
+
+        ddl_statement = "CREATE TABLE {} (atom_id{}, key)".format(
```
dkfellows, line 213: This line looks weird

dkfellows: Did you mean `"CREATE TABLE {} (atom_id, {})"` instead?

Christian-B: Agreed

Christian-B:
Idea from @dkfellows  Rather than creating a keys view
CREATE VIEW IF NOT EXISTS counter AS
WITH RECURSIVE
cnt(x) AS (
SELECT 0
UNION ALL
SELECT x+ ***Sampling_interval*** FROM cnt
LIMIT ***max(key)***/***Sampling_interval***
)
SELECT x FROM cnt;

Christian-B:
idea from @Donal .  Get rid of all the intermediate views and instead have just one complex nested view!

Christian-B:
> 
> 
> Also wondering whether the general recording data type should be an enum.

I was thinking of data_specification.enums.data_type

Christian-B:
Idea from @dkfellows replace table_type "exists" with "events"

dkfellows:
The approach looks pretty reasonable. I'm not sure what the baseline tables should be or what indices are needed, but there certainly looks to need to be one on the timestamp columns, as consumers will often want to `ORDER BY timestamp`. (The timestamps might be time or might be simulation ticks; the difference is irrelevant here.)

dkfellows:
The idea for a `counter` view came from https://stackoverflow.com/a/26241151/301832 with a little massaging.

dkfellows:
Joining the counter (which can be made as long as required; the 100k in the code sample is a sample only) with the underlying state data allows the creation of views where each timestamp has information. This can then be naturally joined with the equivalent for other neuron IDs to get a complete view of all state, including NULLs where information is absent. Being able to order by the timestamp is critical for this to work. Partial recording fixups can be inserted into this system too.

For the event recording, the real question is whether we want to have multiple events per timestep; if so, it might be that the correct way to _eventually_ present the mapping is, for each atom ID, as a relation from timestep to the number of events associated with that atom ID in that timestep. When only zero or one are possible, that's boolean in effect, but at least for poisson sources the number of times they can fire per timestep can be definitely higher than 1.

dkfellows:
If instead we're presenting the data as a mapping from atom ID to timestamp (as is the case in the current data) then there's a need for getting the data for a particular atom ordered by timestamp, which sounds like two indices for fast selection by ID and ordering.

Christian-B:
As for event recording we must support more than one event(spike) per key(timestep) and atom(neuron) id.  Representing this as a count was considered but not done currently as in all known cases the data is returned to the users as multiple rows with the same timestep and id rather than in the timestep, id , count format.

alan-stokes:
alive? dead?

Christian-B:
Now https://github.com/SpiNNakerManchester/sPyNNaker/pull/1051

PR #652, opened 2020-08-26
Provenance database tweaking

dkfellows:
Trying to put more useful info into the provenance database, specifically the coordinates (chip or core) of the origin of the provenance info, where relevant. This extends the `source` table.


dkfellows:
With a bit of playing around, I can use this to extract heat maps like this (done with Google Sheets):
Screenshot 2020-08-26 at 19 51 15
(This is a pair of maps of the number of P2P packets routed off-chip — because I had to pick at least _something_ to plot — showing that in this example at least, we have a lot more traffic close to the ethernet chip. This isn't a very surprising outcome. The low activity of 4,1 is in the data, and the coordinates were all automatically generated.)

andrewgait:
Looks good to me!  Was (5,2) a dead chip on this particular board?

dkfellows:
I don't know; probably? The faults on boards usually seem to be in that sort of area.

PR #655, opened 2020-09-01
wait for interrupt in a common header

dkfellows:
This shouldn't lengthen the code, since it generates the same code as always, yet it allows us to get rid of one of the few calls that remains in our inner loops. (We also retain `spin1_send_mc_packet()` and so on, but those are doing a lot more work.)

### DIFF
--- a/c_common/front_end_common_lib/include/spin1-wfi.h
+++ b/c_common/front_end_common_lib/include/spin1-wfi.h
@@ -27,7 +27,7 @@
 //!     see [the relevant ARM documentation][wfi] (it's hardware magic,
 //!     specific to the ARM968).
 //!
-//! [wfi]: [https://developer.arm.com/documentation/ddi0311/d/system-control-coprocessor/cp15-register-descriptions/cp15-c7-core-control-operations]
+//! [wfi]: https://developer.arm.com/documentation/ddi0311/d/system-control-coprocessor/cp15-register-descriptions/cp15-c7-core-control-operations
 static inline void spin1_wfi(void) {
     register uint32_t value = 0;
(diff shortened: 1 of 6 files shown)


c_common/front_end_common_lib/include/spin1-wfi.h
```diff
@@ -0,0 +1,35 @@
...
+
+/*! \brief Wait for interrupt.
+ *  \details Inline version of code that appears in spin1_api so that we can
+ *      get more compact code. For a description of what this actually does,
+ *      see <a href="https://developer.arm.com/
+documentation/ddi0311/d/system-control-coprocessor/cp15-register-descriptions/
+cp15-c7-core-control-operations">the relevant ARM documentation</a>
+ *      (it's hardware magic, specific to the ARM968).
+ */
+static inline void spin1_wfi(void) {
```
rowleya, line 32: I guess you want this inline; is it worth giving it a different name since it isn't in spin1_api?  It might make people look for it in the wrong place...

rowleya (changes requested):
This looks generally OK; just wondering if calling it spin1_wfi is a good idea given that it isn't part of spin1_api?

c_common/front_end_common_lib/include/spin1-wfi.h
dkfellows: The name is carried forward from previous uses. Suggestions welcome!

rowleya (approved):

c_common/front_end_common_lib/include/spin1-wfi.h
dkfellows: Simple option! `wait_for_interrupt()`

PR #660, opened 2020-09-15
Fix sub timestep STDP issues

rowleya:
This is needed to make a test fail!


rowleya:
Closing as now merged in to #632

PR #661, opened 2020-09-17
Router provenance mapper utility

dkfellows:
This produces heatmap plots for all the router provenance. It can't plot other types of provenance (e.g., anything at core level).

It takes the name of the provenance database, and spits out a whole load of PNG files _with fixed names in the current directory_. Example output:

External_P2P_Packets


dkfellows:
Note also that this requires the additional `seaborn` package to work.

alan-stokes:
total router table entries? unique routes in router after and before compresion.? 

please and thank you

dkfellows:
> total router table entries? unique routes in router after and before compresion.?

Is that information in the database? If so, this can extract it. Just add the relevant key/filename to the list…

dkfellows:
Discovered how to have optional dependencies, so we can have a plotting tool that needs extra libraries without forcing all users to have them. _Documented this in the README._

Gory details of how that works are really at https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies and https://setuptools.readthedocs.io/en/latest/userguide/entry_point.html#dependency-management 
For some reason Google likes to point to the single most useless page in that documentation tree, so use these links instead!

PR #662, opened 2020-09-17
Reallocate spalloc job if known with sick board suffers

Christian-B:
Requires: https://github.com/SpiNNakerManchester/SpiNNMachine/pull/139

TODO!
Report the list of ipaddresses that are sick and not in the list.
Ideally by telling the spalloc server which delists the boards and triggers manual intervention.


spinn_front_end_common/interface/interface_functions/spalloc_machine_generator.py
```diff
@@ -0,0 +1,148 @@
...
+                machine_details, txrx = machine_generator(
+                    hostname, bmp_details, downed_chips, downed_cores,
+                    downed_links, board_version, auto_detect_bmp,
+                    scamp_connection_data, boot_port_num,
+                    reset_machine_on_start_up,
+                    max_sdram_size, repair_machine, ignore_bad_ethernets,
+                    default_report_directory)
+                known = False
+                for sick_ip in sick_boards:
+                    if sick_ip in ex.ipaddress:
```
dkfellows, line 141: This logic seems backwards. We want to exclude known bad boards from reporting, not work through the list of bad boards and check if any of them are in there. Though I suppose recovering if any of them were known is somewhat sensible too.

Need to be clear exactly what conditions trigger reporting and what trigger retrying; they might not be the same!

spinn_front_end_common/interface/interface_functions/spalloc_machine_generator.py
```diff
@@ -0,0 +1,148 @@
...
+                    scamp_connection_data, boot_port_num,
+                    reset_machine_on_start_up, max_sdram_size, repair_machine,
+                    ignore_bad_ethernets, default_report_directory)
+            except SpinnMachineCorruptionException as ex:
+                if n_boards and n_boards > self.MAX_RESTART_SIZE_IN_BOARDS:
+                    raise
+                if n_chips and n_chips > self.MAX_RESTART_SIZE_IN_BOARDS * 48:
+                    raise
+                machine_generator = MachineGenerator()
+                machine_details, txrx = machine_generator(
```
dkfellows, line 132: Definitely should not call this here!

dkfellows (changes requested):

spinn_front_end_common/interface/interface_functions/spalloc_machine_generator.py
dkfellows: The problem is that this is the code that just threw, and the generator's stateless. It's got an _extremely_ high chance to throw again.

Christian-B:
So out of date and old easier to redo if ever needed.

rowleya:
Actually happens on spalloc itself now really anyway!

PR #663, opened 2020-09-18
Get rid of signed/unsigned warnings

dkfellows:
This is done by avoiding bubblesort and using [insertion sort](https://en.wikipedia.org/wiki/Insertion_sort) instead. This should be a bit faster too, and the code is shorter.


rowleya (approved):
Looks like a good idea; approved when it passes

dkfellows:
All passed

PR #664, opened 2020-09-21
Inline all TDMA things

rowleya:
This mostly just moves things into the header for inlining.  It also re-enables the counter for out-of-time TDMA.

### DIFF
--- a/c_common/front_end_common_lib/include/tdma_processing.h
+++ b/c_common/front_end_common_lib/include/tdma_processing.h
@@ -38,9 +38,18 @@ typedef struct tdma_parameters {
     uint32_t time_between_sends;
 } tdma_parameters;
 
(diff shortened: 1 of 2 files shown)


alan-stokes:
this will double the code for SSP and SSA and possibly other things which send packets in different areas of the code

dkfellows (approved):

dkfellows:
> this will double the code for SSP and SSA and possibly other things which send packets in different areas of the code

But are either SSP or SSA so large as for this to be a problem?

alan-stokes:
> > this will double the code for SSP and SSA and possibly other things which send packets in different areas of the code
> 
> But are either SSP or SSA so large as for this to be a problem?

no, but tis always worth knowing the consequences of your actions. otherwise I'm not fussed about this pr.

PR #666, opened 2020-09-29
Run synchronized

rowleya:
This PR adds functionality to allow "closed loop" synchronization with an external agent, such as the HBP NRP.  This adds to the simulation environment by having a defined number of time steps to be run before the simulation pauses and waits for a message from the host to continue.  The check for the simulation to be finished has been encapsulated and combined with this functionality for ease of use.  The simulation still has an overall number of steps to run for, and so nothing is done between the simulation pausing and the message being received, thus speeding up the "closed loop" cycle.


alan-stokes:
why is this work not just using the auto pause and resume functionality to lock the max steps per cycle and using the notification process to get the system to restart?

rowleya:
Because it is too slow.  This requires nothing to be sent from SpiNNaker, just a regular (and either in time or slightly slower time) signal from the host to say "continue".

rowleya:
Correction: any signal that is slower is fine, not just slightly.  In fact you can entirely pause the whole simulation as long as you want,  just by not sending the signal, which is ideal for the NRP integration.

alan-stokes:
can i claim a review then. coz if thats too slow. i want to know how the heck this is doing it "faster"

rowleya:
Yes, please feel free to review.

alan-stokes:
> Correction: any signal that is slower is fine, not just slightly. In fact you can entirely pause the whole simulation as long as you want, just by not sending the signal, which is ideal for the NRP integration.

now youve confused me. if slower is fine. (and given the system will hang forever without the notification response being set. the auot pause and resume func sounds fine..... so confused.

rowleya:
Side note: The NRP is currently being refactored so full testing of this with the NRP will have to wait until that is complete.  This doesn't necessarily stop this being merged as it works in itself, as shown in the new test

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -799,41 +820,33 @@ def _calc_run_time(self, run_time):
...
             total_run_timesteps * machine_time_step_ms *
             self.time_scale_factor)
 
-        # Convert dt into microseconds and divide by
-        # realtime proportion to get hardware timestep
+        # Convert dt into microseconds and multiply by
+        # scale factor to get hardware timestep
         hardware_timestep_us = int(round(
-            float(self.machine_time_step) / float(self.time_scale_factor)))
+            float(self.machine_time_step) * float(self.time_scale_factor)))
```
andrewgait, line 835: Are you sure this is now the right way round?  I don't remember this being wrong previously...

andrewgait (changes requested):
This looks fine to me.  I agree with @alan-stokes that it may be possible to amalgamate this with auto-pause and resume in the future, but for now I'm happy with this.  I just have one minor issue - see comment below.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: I forgot I had changed this, but the timestep on the machine is definitely multiplied by the timescale factor, as otherwise it would get smaller...

andrewgait: Ah, okay.  I forget every so often that the time_scale_factor isn't a fraction but an integer... so happy to merge this now.

andrewgait (approved):

PR #667, opened 2020-09-29
Actually report problems to spalloc

dkfellows:
I'm not yet sure that I've got the logic for this right.


dkfellows:
Most of the changes are just a merge of `master`

dkfellows:
This is now just a broken branch. Yes, we should report problems... but I don't think I can mend this branch to do it any more.

PR #670, opened 2020-10-05
Even Alan can not measure power on virtual chips

Christian-B:


alan-stokes:
requests 1 test which can be ran all the time. which tests this combination of debug mode / power monitors and external devices *sticks tongue out at tester*

alan-stokes (dismissed):
wheres the bespoke test!

PR #671, opened 2020-10-07
Pop based master pop table

rowleya:
Makes bitfields work better with the pop_based_master_pop_table changes in SpiNNakerManchester/sPyNNaker#776.

The changes include:
 - The bit fields are no longer split into three lists, instead having flags in each entry indicating if they are merged and/or redundant
 - Remove some things that are just not used
 - Bit fields filters are stored in memory in key order, so sorting by redundancy is done here instead
 - Check has been added that there is 1 bitfield per processor per routing table entry at most
 - (Minor change) The system execution process has been updated to allow the printing of any errors


c_common/models/compressors/src/sorter/bit_field_reader.h
```diff
@@ -150,6 +150,30 @@ static inline void print_structs(
     }
 }
 
+static uint32_t n_redundant(filter_info_t *filter) {
+    uint32_t n_atoms = filter->n_atoms;
+    uint32_t n_words = get_bit_field_size(n_atoms);
+    return n_atoms - count_bit_field(filter->data, n_words);
+}
+
+//! \brief Sort a subset of the bitfields by the redundancy
```
alan-stokes, line 159: more docs needed. params, etc etc

c_common/models/compressors/src/sorter/bit_field_reader.h
```diff
@@ -150,6 +150,30 @@ static inline void print_structs(
     }
 }
 
+static uint32_t n_redundant(filter_info_t *filter) {
```
alan-stokes, line 153: so this is bemusing. line 39 detect_redundant_packet_count is doing the same thing as this. and line 39 is more doced. but yours is a better impl. i recommend merging them into 1

alan-stokes (changes requested):

spinn_front_end_common/interface/interface_functions/machine_bit_field_router_compressor.py
```diff
@@ -222,19 +222,25 @@ def __call__(
...
+                functools.partial(
+                    self._check_bit_field_router_compressor_for_success,
+                    host_chips=on_host_chips,
+                    sorter_binary_path=bit_field_sorter_executable_path,
+                    prov_data_items=prov_items),
+                [CPUState.FINISHED], True,
+                "bit_field_compressor_on_{}_{}_{}.txt",
+                [bit_field_sorter_executable_path], progress_bar,
+                logger=logger)
+        except SpiNNManCoresNotInStateException as e:
```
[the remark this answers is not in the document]
rowleya: Well, I can add in just "Exception" since I have no idea what else might come out of there...  Note however that I do only want to catch that exception, the others can just raise...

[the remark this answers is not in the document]
rowleya: Hmm, from reading PEP8 and other advice, you really shouldn't catch exceptions that you don't want to handle...

spinn_front_end_common/utilities/system_control_logic.py
```diff
@@ -49,6 +49,9 @@ def run_system_application(
         Or `None` for all binaries
     :param progress_bar: Possible progress bar to update.
            end() will be called after state checked
+    :param logger:
+        If provided and iobuf is extracted, will be used to log errors and
+        warnings
     :type progress_bar: ~spinn_utilities.progress_bar.ProgressBar or None
```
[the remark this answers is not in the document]
rowleya: "Fixed"

c_common/models/compressors/src/sorter/bit_field_reader.h
rowleya: Done

c_common/models/compressors/src/sorter/bit_field_reader.h
rowleya: Merged

c_common/models/compressors/src/bit_field_common/routing_tables.h
```diff
@@ -67,7 +67,8 @@ entry_t* routing_tables_append_get_entry(void) {
     // check that we're not hitting the max entries supported by the table
     if (multi_table.n_entries == (int) multi_table.max_entries) {
         log_error(
-                "there is no more space in this multi-table for this entry.");
+                "there is no more space out of %d entries in this multi-table"
+                "for this entry.", multi_table.max_entries);
```
andrewgait, line 71: I've noticed this more in various places recently - grammatically we should be starting error / log messages such as this with a capital letter.  Probably worth going through all errors / warnings etc. and checking grammar / spelling etc. before we go ahead with a release...

andrewgait (approved):
This looks fine, one minor request.

Also, just recording my dislike for having to use "int" in various places rather than "uint32_t"; I know why it's generally necessary but it would be nice if we could work on sorting this out at some point...

c_common/models/compressors/src/bit_field_common/routing_tables.h
rowleya: Fixed

PR #672, opened 2020-10-09
fix for command sender

alan-stokes:
this fixes external device support so that it actually works again. 

tested by 
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/473


spinn_front_end_common/utility_models/abstract_one_app_one_machine_vertex.py
```diff
@@ -71,14 +71,6 @@ def remember_associated_machine_vertex(self, machine_vertex):
         super(AbstractOneAppOneMachineVertex, self).\
             remember_associated_machine_vertex(machine_vertex)
 
-    @overrides(AbstractGeneratesDataSpecification.generate_data_specification)
```
andrewgait, line 74: flake8 is saying you need to take this out of the imports; while you're there, I think AbstractHasAssociatedBinary can go too (and you can delete get_binary_file_name(...) from above as well).

andrewgait (changes requested):

spinn_front_end_common/utility_models/abstract_one_app_one_machine_vertex.py
alan-stokes: fixed

andrewgait (approved):

PR #673, opened 2020-10-13
Don't error just warn

rowleya:


dkfellows (approved):

andrewgait (approved):
Seems fair enough for now but at some point we should probably revisit whether there are situations where an error is more appropriate than a warning.

dkfellows:
What's the actual minimum inter-send time? The current default (50) smells like it could be arbitrary.

rowleya:
A core needs ~17 cycles for a receive I think + some handling time, so 50(us?) does seem very high; even at 30 clock cycles per receive, you would expect a 150ns (=0.15us) to be more accurate.

dkfellows:
Continue that thought in the linked issue.

alan-stokes:
> A core needs ~17 cycles for a receive I thin

well @lplana was very much on the view that it took 30 ish cycles to get past the interrupt layer in the c code. and the 10 extra was from the ciruclar buffer. 

its also worth noting that the global TDMA came to the same conclusion. so its not exactly arbitary

PR #679, opened 2020-10-21
allows key constriants on vertex to be applied to all outgoing constr…

alan-stokes:
…aints.


Christian-B:
@alan-stokes Check my refactory. If you agree and green Merge it!

PR #680, opened 2020-10-21
Use the AbstractContextManager from SpiNNUtils

dkfellows:
Rather than dealing with the hacky mess that is Python's context manager protocol directly.


Christian-B (approved):
These all look fine to me.

I did not see a Spynakker8 branch to run the integration tests.

Other than the spalloc one which appears to bring in rig these are good to go.

PR #681, opened 2020-10-23
More provenance

dkfellows:
Now that we're looking at actually doing something more useful with provenance (see #661) it would really help if we had better provenance actually coming out of the system. This is PR is about improving all this stuff.

Extends the API provided by `ProvidesProvenanceDataFromMachineImpl` to make it much easier to make a vertex that spits out the standard provenance and its own type-specific fields.


dkfellows:
The key simplification is that, with this change, _most_ vertex types can just `yield` the various `ProvenanceDataItem`s that they want to provide rather than having to do the splicing into the list of provenance being built themselves. This leads to less nested code and is easier to get obviously right. (We can also standardize a bunch of other things too.)

It doesn't work for `PopulationMachineVertex` in sPyNNaker. That (alone of all our vertices) wants to examine the standard provenance as well.

spinn_front_end_common/interface/provenance/provides_provenance_data_from_machine_impl.py
```diff
@@ -210,64 +211,65 @@ def _read_basic_provenance_items(self, provenance_data, placement):
...
     def _get_remaining_provenance_data_items(self, provenance_data):
         """
-        :param list(ProvenanceDataItem) provenance_data:
-        :rtype: list(ProvenanceDataItem)
+        :param list(int) provenance_data:
+        :rtype: list(int)
         """
         return provenance_data[self.NUM_PROVENANCE_DATA_ENTRIES:]
 
+    def _get_extra_provenance_items(
```
Christian-B, line 254: I do not like a function to be overwitten to start with an underscore.

Christian-B: Why is location a tuple when the only place calling it has x, y, p and has to make a tuple and all places that use it split it back out to x, y, p

spinn_front_end_common/interface/provenance/provides_provenance_data_from_machine_impl.py
```diff
@@ -284,4 +286,9 @@ def get_provenance_data_from_machine(self, transceiver, placement):
         """
         provenance_data = self._read_provenance_data(
             transceiver, placement)
-        return self._read_basic_provenance_items(provenance_data, placement)
+        items = self._read_basic_provenance_items(provenance_data, placement)
+        label, x, y, p, names = self._get_placement_details(placement)
```
Christian-B, line 290: Why is self._get_placement_details(placement) done here AND in self._read_basic_provenance_items( which is only called from here?

Why not have read_basic_provenance_item and get_remaining_provenance_data_items take the same parameters?

c_common/models/live_packet_gather/src/live_packet_gather.c
```diff
@@ -33,12 +33,16 @@
...
     //! Count of overflows when no payload was sent
     uint32_t number_of_overflows_no_payload;
     //! Count of overflows when a payload was sent
     uint32_t number_of_overflows_with_payload;
-};
+    //! Number of events gathered and recorded
+    uint32_t number_of_gathered_events;
+    //! Number of messages sent to host
+    uint32_t number_of_sent_messages;
+} provenance_data_struct;
```
Christian-B, line 45: our naming pattern for typdedef to always end the name with _t

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
```diff
@@ -48,6 +48,47 @@ class RouterProvenanceGatherer(object):
         # transceiver
         '_txrx',
     ]
+    __DROPPED_MC_MSG = (
```
Christian-B, line 51: I highly dislike the moving of stings only used on one place to constants.

Especially when they need formatting.

Why not keep them in place with f" style formatting.

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
```diff
@@ -135,288 +180,236 @@ def _add_router_provenance_data(
...
-
-    @staticmethod
-    def _has_virtual_chip_connected(machine, x, y):
+            message=self.__DROPPED_MC_MSG.format(
+                x, y, diagnostics.n_dropped_multicast_packets))
+        yield ProvenanceDataItem(
+            names + ["Dropped_Multicast_Packets_via_local_transmission"],
+            diagnostics.user_3,
+            report=(diagnostics.user_3 > 0),
+            message=self.__DROPPED_LOCAL_MC_MSG.format(
```
Christian-B, line 321: I know this method is very long but moving the Strings out only makes it even harder to read!

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
```diff
@@ -88,23 +129,28 @@ def __call__(
...
+    def __summary_items(self):
+        """
+        :rtype: iterable(ProvenanceDataItem)
+        """
+        yield ProvenanceDataItem(
             ["router_provenance", "total_multi_cast_sent_packets"],
-            self._total_sent_packets))
-        prov_items.append(ProvenanceDataItem(
+            self._total_sent_packets)
+        yield ProvenanceDataItem(
```
Christian-B, line 142: using yield instead of building lists to only use in an extend call is APPROVED!

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
```diff
@@ -135,288 +180,236 @@ def _add_router_provenance_data(
...
         """
-        :param ~.RouterDiagnostics router_diagnostic:
-        :param ReInjectionStatus reinjection_status:
+        :param dict(tuple(int,int),ReInjectionStatus) reinjection_data:
+        :param int x:
+        :param int y:
+        :rtype: ReInjectionStatus or None
+        """
+        status = None
+        if reinjection_data is not None:
```
Christian-B, line 248: why have a temp status instead of just two returns?

Christian-B (changes requested):
Will need a second pass but here are some early suggestions

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
dkfellows: You'd rather it was the other way? OK. Can do.

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
dkfellows: Fix for one of these is fix for all.

c_common/models/live_packet_gather/src/live_packet_gather.c
dkfellows: Experience with doxygen suggests:
1. Make the type name unique in the codebase (because otherwise you get weird crosslink errors).
2. Make a typedef with the same name as the struct (because otherwise you will get very confused).

I've gone with that, and I've used the `_t` suffix instead of the ugly `_struct`.

spinn_front_end_common/interface/provenance/provides_provenance_data_from_machine_impl.py
dkfellows: They only split it out so they can use it (in inconsistent ways!) in messages that are only parsed by humans, not machines.

dkfellows:
I think there's an opportunity to standardise how we _describe_ cores to users in provenance messaging. I suggest always using:
```python
std_descriptive_label = f"{label} on {x},{y},{p}"
```
because that's approximately what is most commonly used and definitely always produces something that they can both relate to (the label) and debug with (the location).

Of course, this is a more intrusive change.

spinn_front_end_common/interface/interface_functions/router_provenance_gatherer.py
dkfellows: No idea what I was thinking there.

spinn_front_end_common/interface/provenance/provides_provenance_data_from_machine_impl.py
dkfellows: That puzzled me too.

spinn_front_end_common/interface/provenance/provides_provenance_data_from_machine_impl.py
dkfellows: Method is now called `parse_extra_provenance_items`.

Christian-B (approved):
Looks ok to me.
Glad you have an Integration test branch as I did not check all the fine details

PR #685, opened 2020-10-27
Minor changes to the OneAppOneMachine Vertex

Christian-B:


alan-stokes:
assuming passes. should be rubbered. silly travis being slow: using this as a marker for refinding

PR #686, opened 2020-10-28
check machine level first

Christian-B:
depends on https://github.com/SpiNNakerManchester/PACMAN/pull/318

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/process_partition_constraints.py
+++ b/spinn_front_end_common/interface/interface_functions/process_partition_constraints.py
@@ -25,42 +25,27 @@ class ProcessPartitionConstraints(object):
         partition request it.
 
     :param ~pacman.model.graphs.machine.MachineGraph machine_graph:
-    :param ~pacman.model.graphs.application.ApplicationGraph application_graph:
     """
 
     def __call__(self, machine_graph):
         """
         :param ~.MachineGraph machine_graph:
-        :param ~.ApplicationGraph application_graph:
         """
-        if machine_graph.application_level_used:
-            # generate progress bar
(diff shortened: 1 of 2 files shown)


spinn_front_end_common/interface/interface_functions/process_partition_constraints.py
```diff
@@ -28,12 +28,12 @@ class ProcessPartitionConstraints(object):
...
     """
 
-    def __call__(self, machine_graph, application_graph=None):
+    def __call__(self, machine_graph):
         """
         :param ~.MachineGraph machine_graph:
         :param ~.ApplicationGraph application_graph:
         """
-        if application_graph is not None:
+        if machine_graph.application_level_used:
```
rowleya, line 36: Looking at this algorithm in particular, you probably don't even need to check; the code below checks the machine level and then only checks the application level 1) if the machine level doesn't have the interface and 2) if the machine level has an app level.  So this check, and the function "process_machine_partition" are both probably not needed...

Christian-B: That was a choice between s slight speed increase vs algorithm complexity.

I was unsure and the their is a clear opinion for KISS, So will change

rowleya: Yep, that looks better :)

rowleya (approved):
Oh the simplicity!

alan-stokes:
assuming passes. should be rubbered. silly travis being slow: using this as a marker for refinding

alan-stokes:
pacman version killed.

PR #690, opened 2020-11-03
Fixes for the Python docs

dkfellows:
Only really needed proper work in one package (`…on_chip_router_table_compression`) plus a few tweaks in the config.


rowleya (approved):
Looks good to me

PR #691, opened 2020-11-03
adjust api accordingly.

alan-stokes:
needed to support machine level api usage with app level adjustments

### DIFF
--- a/spinn_front_end_common/abstract_models/abstract_rewrites_data_specification.py
+++ b/spinn_front_end_common/abstract_models/abstract_rewrites_data_specification.py
@@ -33,7 +33,7 @@ class AbstractRewritesDataSpecification(object):
         "machine vertex, the SpiNNFrontEndCommon function DSGRegionReloader "
         "will not check this vertex")
 
-    def __new__(cls, *more):
+    def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


Christian-B (dismissed):
AbstractRewritesDataSpecification init needs to be removed as it is now also used as a Machine Vertex

spinn_front_end_common/abstract_models/abstract_rewrites_data_specification.py
```diff
@@ -23,7 +26,18 @@ class AbstractRewritesDataSpecification(object):
...
 
-    __slots__ = ()
+    __slots__ = []
+
+    _WRONG_VERTEX_TYPE_ERROR = (
+        "The vertex {} is not of type MachineVertex. By not being a "
+        "machine vertex, the SpiNNFrontEndCommon function DSGRegionReloader "
+        "will not check this vertex")
+
+    def __new__(cls, *more):
```
Christian-B, line 36: This should be def __new__(cls, *args, **kwargs):

Then it safely covers all cases,

While kwargs is not the spinnaker way of working it is correct in this case

alan-stokes: fixed

PR #693, opened 2020-11-05
Add C actions

andrewgait:
Add C actions workflow

### DIFF
--- a/.github/workflows/c_actions.yml
+++ b/.github/workflows/c_actions.yml
@@ -82,7 +82,8 @@ jobs:
         
     - name: Lint C code using Vera++
       run: |
-        support/run-vera.sh c_common/front_end_common_lib
+        support/run-vera.sh c_common/front_end_common_lib/include
+        support/run-vera.sh c_common/front_end_common_lib/src
         support/run-vera.sh c_common/models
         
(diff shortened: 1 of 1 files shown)


.github/workflows/c_actions.yml
```diff
@@ -44,11 +44,6 @@ jobs:
         repository: SpiNNakerManchester/SupportScripts
         path: support
         
-    - name: Set GITHUB_BRANCH
```
andrewgait: I would have thought GITHUB_BRANCH would be needed here, if there are branches with the same name that need to be checked out from spinn_common or spinnaker_tools?

Christian-B: GITHUB_BRANCH was needed when we where doing pull requests as then the environment variable GITHUB_REF was not usable.

As we are only doing push ones clone2.sh directly uses GITHUB_REF so GITHUB_BRANCH is not longer needed to get any downstream branches with the same name

andrewgait: Ah, okay - good point.  Needs removing in other C actions branches as well in that case... I'll do that.

dkfellows (approved):

PR #697, opened 2020-11-09
Paper tests

alan-stokes:
this adds functionality for tracking values for the speed up paper. probably needs a mega clean up


dkfellows:
We will not be committing this work. This branch will be left in place.

PR #702, opened 2020-11-17
Fix -Wshadow compiler warnings

andrewgait:
Fixing some compiler warnings I just noticed - code that was there was a bit dangerous, shall we say...

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/500


rowleya (approved):
Looks like some oopses in there now fixed!

alan-stokes:
yep. looks fine

alan-stokes (approved):

PR #703, opened 2020-11-17
Enable -Wold-style-definition and fix fault it finds

dkfellows:
Old-style function definitions (with `foo_bar()` instead of `foo_bar(void)`) allow _any_ number of arguments rather than _none,_ and are always wrong in our code. This PR makes them much easier to pick up.


dkfellows:
The generated code should be identical to before (except in cases where things are actually broken and we want the build to fail) so I don't plan to rerun the integration tests for this one.

andrewgait (approved):
Looks good to me - thanks for spotting it.

PR #704, opened 2020-11-20
Enable specific documenting of __call__() methods

dkfellows:
See also SpiNNakerManchester/sPyNNaker#933 (but this is independent)


spinn_front_end_common/abstract_models/impl/tdma_aware_application_vertex.py
```diff
@@ -44,12 +45,13 @@ class TDMAAwareApplicationVertex(ApplicationVertex):
         "The TDMA fell behind by {} times on core {}, {}, {}. "
         "try increasing the time_between_cores in the corresponding .cfg")
 
-    def __init__(self, label, constraints, max_atoms_per_core):
+    def __init__(self, label=None, constraints=None,
```
Christian-B, line 48: I disagree with adding default values here!   That is the role of the super classes.

Christian-B (commented):
Look good except for adding the default to TDMA.

I assume of course sphinx does the docs correctly.

spinn_front_end_common/abstract_models/impl/tdma_aware_application_vertex.py
dkfellows: Removed

PR #706, opened 2020-11-25
Fix continue

rowleya:
Fixes a couple of things that make the integration with the NRP smoother.  In particular:
  - Remove some log output that fills the logs eventually
  - Fix "run forever" to have the correct final state when completed
  - Fix "run forever" to not do data extraction at the start of the run(!)
  - On stopping an application, send signals to avoid issues when the system is paused.

Tested by: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/fix_continue/4/pipeline/


andrewgait (approved):
Looks fine to me, will merge it.

PR #707, opened 2020-11-25
clone partitions when building new machine graphs

Christian-B:
As partition's can now only be added to one graph we have to clone then when copying graphs.

Also now ASB _run will return immediately if both graphs are empty.

Depends on https://github.com/SpiNNakerManchester/PACMAN/pull/331
Tested by 
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/158
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/506

AbstractSpinnakerBase._build_graphs_for_usage just calls clone methods on graph

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -761,7 +761,8 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
         self._application_graph = self._original_application_graph.clone()
 
         # sort out machine graph
-        self._machine_graph = self._original_machine_graph.clone()
+        self._machine_graph = self._original_machine_graph.clone(
+            self._application_graph)
 
     def __timesteps(self, time_in_ms):
         """ Get a number of timesteps for a given time in milliseconds.
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -741,6 +741,27 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
             raise ConfigurationException(
                 "Illegal call after simulation is shutdown")
 
+    def _should_run(self):
+        """
+        Checks if the simulation should run.
+
+        Will warn the user if there is no need to run
+
+        :return: True if and only if one of the graphs has vertices in it
+        :raises ConfigurationException: If the current state does not
+            support a new run call
+        """
+        self.verify_not_running()
+
+        if self._original_application_graph.n_vertices:
+            return True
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1662,8 +1652,7 @@ def _do_mapping(self, run_time, total_run_time):
 
         # only add the partitioner if there isn't already a machine graph
         algorithms.append("MallocBasedChipIDAllocator")
-        if (self._application_graph.n_vertices and
-                not self._machine_graph.n_vertices):
+        if self._application_graph.n_vertices:
```
rowleya, line 1655: The comment above this says to only add the partitioner if there isn't already a machine graph.  I am guessing this is from the "re-run" mode of operation i.e. a run has already happened, the application graph hasn't changed and so the machine graph doesn't need to be re-created.  Removing this check might mean that a re-run causes a re-partition.

rowleya (changes requested):
Looks like a major improvement in terms of simplification.  One comment made to be checked before merging.

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: Yes. Changed the check to just if not self._machine_graph.n_vertices:
We know that from should_run if the machine_graph has not vertexes the application one must.

Christian-B:
One thing I discovered is that GFE exposed the application and machine graphs after run.

If the user then adds directly to these graphs all bets are off.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Should_run checks the original graphs though, not the clones.  This specific check is testing if partitioning should happen; it shouldn't if the graphs haven't changed and you are re-running.

rowleya: I might be getting the path wrong here though; this is in _do_mapping, so if this has already decided to do all mapping, maybe redoing partitioning here is correct?

Christian-B: The check is now:
if not self._machine_graph.n_vertices:

This confirms if partitioning is needed or not,
The second check self._application_graph.n_vertices was to avoid partitioning and empty graph. Should run has already taken care of the empty graphs.

rowleya: OK, sounds good

rowleya (approved):

rowleya:
It looks like the sPyNNaker8 tests are failing just now...

PR #710, opened 2020-11-26
InsertEdgesToExtraMonitorFunctionality cleanup

Christian-B:
As the only machine vertex an ExtraMonitorSupport can have is a ExtraMonitorSupportMachineVertex there is no need to call 
self._process_app_graph_vertex twice with the same vertex


alan-stokes:
failing tests. but i approve when its passing

PR #711, opened 2020-11-26
Convert workflows to use common actions

dkfellows:
See SpiNNakerManchester/SupportScripts#25


dkfellows:
Do not commit until we've merged the upstream `support` and fixed the checkout ref

PR #727, opened 2020-12-15
log warning to file

Christian-B:
cfg setting 
[Logging] warnings_at_end_to_file

if set to true will add warning and above to a warning_logs.txt file,
Regards of what level logging is set to (also in configs)

If true only print a message at end pointing to file
If false repeats warning at end of run.

Depends on: https://github.com/SpiNNakerManchester/SpiNNUtils/pull/128


spinn_front_end_common/interface/spinnaker.cfg
```diff
@@ -32,6 +32,13 @@ warning =
 error =
 critical =
 
+# Note log levels set above does not change the log message reported at the
+# end of the run
+
+# Set to true and all log messages warning of above will be written to file
```
andrewgait, line 38: "warning or above"

andrewgait (changes requested):
Looks good, just a minor edit

spinn_front_end_common/interface/spinnaker.cfg
Christian-B: fixed

andrewgait (approved):

PR #730, opened 2020-12-16
Remove circular dependency

rowleya:
This ended up being a rather large chain of circular dependency removal.  The main headlines:
 - Made the extraction of iobuf a class by itself, which the algorithm then can use, but so can other things.
 - Moved the emergency recovery out of helpful_functions.
 - Removed a pointless function that didn't help much other than creating a circular dependency!

Tested by: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/activity?branch=remove_circular_dependency


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -102,6 +104,21 @@
 _PREALLOC_NAME = 'MemoryPreAllocatedResources'
 
 
+def get_front_end_common_pacman_xml_paths():
```
Christian-B, line 107: I do not like this approach at all!

Would it not be better to have both interface_functions and report_functions each report their own xml file.

Dont have this method at all!

Instead have 
_create_xml_paths
call each of the two functions and directly use the result in 
xml_paths.extend(

Christian-B (dismissed):
Put getting the xml file near the xml files.

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: Note: I believe this has simply been moved to avoid a circular dependency

Christian-B:
Good to merge in my opinion.

PR #733, opened 2021-01-13
remove MundyOnChipRouterCompression

Christian-B:
Leaves behind just a NotImplementedError telling them to change the cfg.

Must be done at the same time or after https://github.com/SpiNNakerManchester/sPyNNaker8/pull/527


Christian-B:
It needs to be pointed out that the issue that triggered the plan to remove the Mundy aplx was NOT caused by the Mundy aplx.

So as yet we have no case where the Mundy aplx has been shown to not work.  That said we also have no case which shows it does any better than the Unordered Compressor. And may case where the Pair compressor was better than either.

rowleya (approved):
This is fine if we are removing this

PR #735, opened 2021-01-18
use config.get_str_list to remove whitespace

Christian-B:
depends on: https://github.com/SpiNNakerManchester/SpiNNUtils/pull/130

Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/344

PR #736, opened 2021-01-20
Several minor fixes to Compressors

Christian-B:
Part of https://github.com/SpiNNakerManchester/PACMAN/pull/347

Builds on SpiNNakerManchester/SpiNNFrontEndCommon#735

BitFieldCompressorReport:
     Reports the bitfield count even if a none bitfield compressor used.
     Does not report chips with no bitfields
     can be turned on with write_bit_field_compressor_report cfg flag

Ability to reduce the time Bitfields compressors take BUT at the cost of probably less that optimal bit fields merged in! 
    cfg param router_table_compression_with_bit_field_retry_count
          default (none) as much as possible
          0: single pass but can loose up to 10% of merged bitfields
          at about 10 the compressor will take twice as long as 0 but less than half the error

Which compressor and retry count added to progress bar so clearly visible to user

Remove confusing log error messages if using MachineBitFieldPairRouterCompressor
 
Moves router_table_compression... param handling to this level.

### DIFF
--- a/c_common/models/compressors/Makefile
+++ b/c_common/models/compressors/Makefile
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 APPS = sorter.mk \
-       bit_field_unordered_compressor.mk \
+       bit_field_ordered_covering_compressor.mk \
        bit_field_pair_compressor.mk \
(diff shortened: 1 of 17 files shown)


alan-stokes:
approved

PR #737, opened 2021-01-22
move generate_router_compression_with_bitfield_report

Christian-B:
Move the param to where GFE can find it

The other half of this PR but less critical is https://github.com/SpiNNakerManchester/sPyNNaker/pull/969


Christian-B:
Actually until GFE has bitfield this is the wrong fix

PR #738, opened 2021-01-25
Remove Python 2.7 support

dkfellows:
PyPI and pip are forcing the issue; we can't keep going much longer.
SpiNNakerManchester/SpiNNUtils#132


dkfellows:
Should I move `require_subclass` into SpiNNUtilities?

Christian-B:
> Should I move `require_subclass` into SpiNNUtilities?

Where is this code?

dkfellows:
Moved.

spinn_front_end_common/abstract_models/abstract_provides_incoming_partition_constraints.py
```diff
@@ -13,12 +13,14 @@
...
 
-from six import add_metaclass
 from spinn_utilities.abstract_base import AbstractBase, abstractmethod
+from spinn_utilities.require_subclass import require_subclass
+from pacman.model.graphs import AbstractVertex
 
 
-@add_metaclass(AbstractBase)
-class AbstractProvidesIncomingPartitionConstraints(object):
+@require_subclass(AbstractVertex)
```
alan-stokes, line 21: might want to chekc the interface function for using those. i think its only machine level these days

alan-stokes: acutally, seems we still support both. maybe we should only support it on 1 though? food for thought

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
```diff
@@ -669,12 +666,8 @@ def send_data_into_spinnaker(
             original_data = bytes(data[offset:n_bytes + offset])
             verified_data = bytes(transceiver.read_memory(
                 x, y, base_address, n_bytes))
-            if PY2:
-                self.__verify_sent_data_py2(
-                    original_data, verified_data, x, y, base_address, n_bytes)
-            else:
-                self.__verify_sent_data_py3(
-                    original_data, verified_data, x, y, base_address, n_bytes)
+            self.__verify_sent_data_py3(
```
alan-stokes, line 669: maybe remove the py3 these days?

PR #742, opened 2021-02-03
adds dial to shut up waiting report

alan-stokes:
tested by: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/539


rowleya (approved):
Again, thought this might go to the algorithms where the waiting is done, but this does make it global I guess

alan-stokes:
you wanted a local control?! thats a much more invasive change, i guess i could allow the function to override the cfg if you want. i thought global would be better than local decisions.

rowleya:
Expected rather than wanted (how I would likely have done it).  But this is fine.

PR #745, opened 2021-02-05
fix write_bit_field_compressor_report

Christian-B:
Also needs https://github.com/SpiNNakerManchester/sPyNNaker/pull/987


rowleya (approved):
I think this looks OK

PR #746, opened 2021-02-08
stinky numpy

alan-stokes:
vom that you need to do this. but meh


Christian-B:
Thanks. While correct
I think this is a sticky plaster at the wrong point.

Christian-B:
I think https://github.com/SpiNNakerManchester/sPyNNaker/pull/990 is a better fix and it addresses the route cause.

alan-stokes:
> Thanks. While correct
> I think this is a sticky plaster at the wrong point.

definitely was a sticky plaster. i persume its safe to delete the brnahces then?

PR #751, opened 2021-02-15
Prealloc cleanup

dkfellows:
We were not allocating a preallocated resource container ahead of time because we were allocating it… with the _wrong type name._ This fixes that and updates the code to work with the (now valid!) assumption that the thing is there.


rowleya (approved):
Looks ok

PR #755, opened 2021-02-16
remove leftover six references

Christian-B:


dkfellows (approved):
Very much so in this one!

PR #741, opened 2021-02-01
Tdma auto add on

alan-stokes:
allows time scale factor to be adjusted for. 

tested by: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/538

PR #756, opened 2021-02-18
Lost spike issues

rowleya:
This helps with issues seen in recent master where the microcircuit couldn't run at 40x slowdown as it could before.  The changes here are to do with the TDMA configuration.  Note that the changes include those from #741, though some changes have been made to this here as well. 

The changes include:
1. The TDMA calculations are now done on clock cycles rather than fractions.  It is now possible to specify the clock cycles for each parameter rather than just the fraction, though that is also supported (testing is done to ensure both are not specified at once).  It turns out it is more useful to specify some things in clock cycles and some things in fractions.
2. The automatic part that calculates the time between spikes is now done on an app-vertex-by-app-vertex manner.  This is needed because some vertices send more spikes than others.  If you take the maximum for all, those vertices which send less may end up sending lots of spikes at the start of a timestep rather than spreading them out.
3. The initial offset between vertices now has its own clock_cycles / fraction options in the config (required because of the above, but also useful in its own right).

This has matching branches in SpiNNMan, PACMAN and spinnaker_tools, but doesn't strictly depend on them.

SpiNNakerManchester/sPyNNaker#1005 depends on this.


spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
```diff
@@ -118,69 +106,133 @@ def __call__(
...
+        # easier bool compares
+        core_set = clocks_between_cores is not None
+        app_set = app_machine_quantity is not None
+
+        # adjust time between cores to fit time scale
+        if not core_set and app_set:
+            n_slots = int(math.ceil(n_cores / app_machine_quantity))
+            clocks_per_phase = (
+                int(math.ceil(overall_clocks_available / n_phases)))
+            # NOTE the plus 1 ensures the last core finishes, if its the worst
```
andrewgait, line 188: "it's" not "its" here

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
```diff
@@ -118,69 +106,133 @@ def __call__(
...
+            # in terms of n keys to transmit
+            clocks_between_cores = clocks_per_phase / (n_slots + 1)
+            logger.debug("adjusted clocks between cores is {}".format(
```
andrewgait, line 191: Capital at start of comment.

andrewgait (changes requested):
Looks fine to me, just a couple of minor edits.

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
rowleya: Fixed

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
rowleya: Fixed I think (along with a few others in the same method)

andrewgait (approved):

PR #762, opened 2021-02-23
More exposed via global variables

Christian-B:
In our current setup the only way to get provenance paths was to access the underscore parameters on the simulator.

I choose not to expose these as formal parameters/methods as that would add more none PyNN syntax

Adding these to Global Variables also made it possible for the last paths used to still be available after sim.end was called.

Same for run_report_directory which gets simulator._report_default_directory

--
In the current setup the configs are also only available between sim.setup and sim.end.

I have added a function to allow a preview of configs to be available before setup. 
For example this could be used to find the board version number
With Spynnaker and GFE each reading in the appropriate cfg file.
With a clear log warning that this is not recommended

The actual configs used are now also available after sim.end for any use.


rowleya (approved):
Looks good

PR #768, opened 2021-03-04
Simplify recording

rowleya:
Remove both the ability to empty recorded buffers live, and the ability to use DMAs for recordings.  The former is never used any more anyway (since auto pause and resume works better for this) and the latter was used but had little effect on performance and will get in the way of other DMA things upcoming (namely those in the split synapse neuron models).

Depended on by:
  - SpiNNakerManchester/sPyNNaker#1032
  - SpiNNakerManchester/SpiNNakerGraphFrontEnd#182
  - SpiNNakerManchester/JavaSpiNNaker#279


andrewgait (approved):
Looks good to me.  Great to be able to reclaim a decent chunk of ITCM back by doing this as well.

dkfellows (approved):

PR #770, opened 2021-03-22
dont fail in error support

Christian-B:
Issues discovered while playing with config files that had an invalid machine.

This caused the algorithms to raise an exception before a ReportFolder was created.

Which in turn cause TagsFromMachineReport to fail hiding the real issue.

Fixes.
1. Do not run TagsFromMachineReport if you dont have a report folder.
2. If you run TagsFromMachineReport dont lets its exception be the main one

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1180,11 +1180,18 @@ class AbstractSpinnakerBase(ConfigHandler, SimulatorInterface):
             self._machine_allocation_controller = executor.get_item(
                 "MachineAllocationController")
             report_folder = executor.get_item("ReportFolder")
-            TagsFromMachineReport()(report_folder, self._txrx)
+            try:
+                if report_folder:
+                    TagsFromMachineReport()(report_folder, self._txrx)
+            except Exception as e2:
+                logger.warning(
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1180,11 +1180,18 @@ def _run_algorithms(
...
+                logger.warning(
+                    "problem with TagsFromMachineReport {}".format(e2),
+                    exc_info=True)
             try:
                 self._shutdown()
-            except Exception:
-                logger.warning("problem when shutting down", exc_info=True)
+            except Exception as e3:
+                logger.warning("problem when shutting down {}".format(e3),
+                               exc_info=True)
```
rowleya, line 1194: Technically, I believe the exc_info=True means that the exception details are displayed anyway, but this isn't critical.

rowleya (approved):
Fine.  One minor comment, but happy to merge without changing this.

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: This way also adds it to the log message repeat at the very end.

rowleya: OK, fair enough

PR #771, opened 2021-03-22
Bits and pieces of code improvement

dkfellows:
* More contexts
* Splitting out shared constants
* Sharing the basic bits and pieces of SQLite interfacing


spinn_front_end_common/utilities/energy_constants.py
```diff
@@ -0,0 +1,51 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Miscellaneous constants used in energy consumption calculations
+"""
+
+#: given from Indar's measurements
```
Christian-B, line 20: Why not just add these to constants.py?

dkfellows:
I think I'll be splitting bits and pieces out of this branch for some time.

PR #774, opened 2021-03-24
Expose report_simulation_top_directory Recorder database

Christian-B:
As the recorder database will have data from multiple segments it needs to go to the top level of the reports

This exposes that in global variables.


dkfellows:
This will need at least one test. Principle seems OK

Christian-B:
now in views

PR #778, opened 2021-03-31
Really want to get strings back in this case

dkfellows:
fixes SpiNNakerManchester/sPyNNaker#1053


andrewgait (approved):
Fixes broken test (locally).  Good to merge once integration tests pass.

PR #799, opened 2021-04-27
Fix extra monitor memory leak

### LINKED ISSUE #796: Extra monitor core runs out of SDRAM on download
The extra monitor core ran out of memory apparently.  It would be hoped that this wouldn't happen.  It would be good to work out why it does.

[error.log](https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/files/6377280/error.log)

rowleya:
Fixes #796
Tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/activity?branch=fix_extra_monitor_memory_leak

### DIFF
--- a/c_common/models/system_models/src/extra_monitor_support.c
+++ b/c_common/models/system_models/src/extra_monitor_support.c
@@ -859,8 +859,8 @@ static void reinjection_read_packet_types(const reinject_config_t *config) {
 
     io_printf(
         IO_BUF,
-        "Setting reinject mc to %d\n Setting reinject pp to %d\n"
-        "Setting reinject fr to %d\n Setting reinject nn to %d\n",
+        "Setting reinject mc to %d\nSetting reinject pp to %d\n"
+        "Setting reinject fr to %d\nSetting reinject nn to %d\n",
         reinject_mc, reinject_pp, reinject_fr, reinject_nn);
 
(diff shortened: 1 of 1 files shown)


rowleya:
I don't know for sure that this fixes the leak, but it does make sure that a free happens where it wasn't before (and the pointer was lost), and the tests pass...

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -1818,7 +1823,9 @@ static void data_out_speed_up_command(sdp_msg_pure_data *msg) {
                     data_out_transaction_id, message->transaction_id);
             return;
         }
+#if 0
```
Christian-B, line 1826: Why do if #0
Would it not be better to either of
1. Just comment out
2. And if def with a flag that could be set at compile time
3. Just delete they easy enough to add back in as as when required.

Christian-B (approved):
Aprroved with or without the ugly if #0

c_common/models/system_models/src/extra_monitor_support.c
rowleya: Agreed, but will leave for a later PR to do this properly with log_debug etc.

PR #815, opened 2021-05-25
Split synapse neuron models

rowleya:
An unfortunate rag-tag of things that helped execute the cortical column in real-time!  Here is a summary:
 - Adds "user callback queue overloads" provenance data (useful for diagnostics)
 - Use "direct sending and waiting" for multicast packets.  This avoids callbacks when lots of packets are sent, and gives back more control to the TDMA.
 - Minor fix to wait_for_interrupt to match that in sark (wasn't working 100% without this).
 - Make it possible to not register a DMA callback using a -2 priority.
 - Make error recovery work more reliably (sleep between provenance updates, don't try to get provenance from RTE cores)
 - Allow SDRAM Allocation to work on virtual machines(!) - this makes it possible to test DSG on things that use this without a real machine
 - Check region sizes when MultiRegionSDRAM is used and warn the user if the resources and the allocation don't match.
 - Allow references to regions and resolve during execution (see also SpiNNakerManchester/DataSpecification#108)


spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
```diff
@@ -182,9 +182,7 @@ def _auto_config_times(
             n_slots = int(math.ceil(n_cores / app_machine_quantity))
             clocks_per_phase = (
                 int(math.ceil(overall_clocks_available / n_phases)))
-            # Note: The plus 1 ensures the last core finishes, if it's the
-            # worst in terms of n keys to transmit
-            clocks_between_cores = clocks_per_phase / (n_slots + 1)
+            clocks_between_cores = clocks_per_phase / n_slots
```
andrewgait, line 185: Was the previous comment / code here incorrect, and the last core can never be the worst in term of n keys to transmit?

andrewgait (changes requested):
Looks good to me - split models were always going to require changes across the board.  I have just one question about one of the changes.

spinn_front_end_common/interface/interface_functions/local_tdma_builder.py
rowleya: The issue is that this is "just in case" rather than being anything else.  So it gives you one extra slot, but at the cost that all slots are smaller.  This ends up with things overrunning slots anyway where they might have been OK in the usual case.  It also complicates things as it adds a slot behind the scenes, meaning that your own manual calculations end up out, and you then have to work out why... it took me a while to find this and fix it!  It is worth noting that much of the code in these branches now doesn't use this at all for neurons, which appears to actually work better!  I believe that this was mostly of concern for Poisson sources (where it is still used I think) but neuron models tend to sort themselves out anyway.  Testing has shown that you don't seem to lose spikes without this in place in any case.

andrewgait (approved):

PR #821, opened 2021-06-03
Add tolerance of 1 core in Json comparison

Christian-B:
Spin4 appears to have recovered some cores,

Adapted the json check to allow for a difference of 1 core from the previously recorded.


Christian-B:
Only a test and master broken so merging

PR #828, opened 2021-06-10
Unittest setup

Christian-B:
See https://github.com/SpiNNakerManchester/SpiNNUtils/pull/159

Plus:
Empty test nuked


rowleya (approved):
Minor comment - the unset_simulator function could be renamed to something like setup_simulator_unittest to be absolutely clear it shouldn't be called from outside.  Non-critical.

Christian-B:
Agreed. While the  unset_simulator function pre existed this change its use has changed over time from needed everywhere to only being needed in unittests

I am going to call it just setup_for_unittest as it does not really setup a simulator just remove any previous ones.  In a followup pr this will also clear any temp dirs setup (as there will be no simulator with report dirs)

PR #835, opened 2021-07-14
Token loop fix

Christian-B:
This PR removes an edge case and fixes a loop

ProcessorToAppDataBaseAddress was both an input and an output in two algorithms.
Now neither inputs it and the two outputs are given different names.
The consumer now uses both

Removed the token loop:
ChipIOBufClearer had an input ReadIOBuf, ChipIOBufExtractor had an input ApplicationRun and an has an input ClearedIOBuf

I choose the order  ClearedIOBuf, ApplicationRu, ChipIOBufExtractor, for two reasons.
1. It leaves the iobuff on core for any advanced debugging
2. Another pr will make ClearedIOBuf, happen on additional runs only

This PR is required for https://github.com/SpiNNakerManchester/PACMAN/pull/405 but can be committed first

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/front_end_common_interface_functions.xml
+++ b/spinn_front_end_common/interface/interface_functions/front_end_common_interface_functions.xml
@@ -766,10 +766,6 @@
                 <param_type>JavaCaller</param_type>
             </parameter>
             <parameter>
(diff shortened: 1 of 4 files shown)


Christian-B:
This will be easier to redo in the no executor work than to merge

PR #839, opened 2021-07-23
Tidy up debug code

dkfellows:
This (parallelling SpiNNakerManchester/sPyNNaker#1103) reduces the impact of messages put in FEC's C code for debugging purposes. It can probably be reduced further, but more eyes make that easier.


c_common/models/compressors/src/bit_field_compressor.c
```diff
@@ -65,8 +65,8 @@ compressor_states previous_compressor_state = UNUSED_CORE;
 //! SDRAM are used for communication between sorter and THIS compressor
 comms_sdram_t *restrict comms_sdram;
 
-// DEBUG stuff
-#if 0
+// DEBUG stuff please leave
+#ifdef DEBUG_COMPRESSOR
```
andrewgait, line 69: Is DEBUG_COMPRESSOR defined anywhere?  The only debug flags I can see defined are FEC_DEBUG and PRINT_DEBUG.  (I guess the user can choose to turn this on from their command line "make" but it would probably make sense in that instance to have an action build where this and similar other options are turned on).

c_common/models/system_models/src/extra_monitor_support.c
```diff
@@ -907,8 +906,8 @@ static inline void reinjection_set_wait2_timeout(uint payload) {
 //! response
 //! \return The payload size of the response message.
 static inline int reinjection_set_timeout_sdp(sdp_msg_t *msg) {
-#if 0
-    io_printf(IO_BUF, "setting router timeouts via sdp\n");
+#ifdef DEBUG_REINJECTOR
```
andrewgait, line 909: Again here I don't remember DEBUG_REINJECTOR being defined anywhere, though again I guess the user can choose to turn this on.  We should probably have an action build which turns these on at least to check the code inside them can be compiled, though.

andrewgait (approved):
Looks good to me, general approval, just a couple of questions relating to debug flags in certain models.

c_common/models/compressors/src/bit_field_compressor.c
dkfellows: No, it's not defined. I suggest that these are a group of messages that it makes sense to review for possible deletion as a set. There's three such groups in the extra monitor. Suggesting that _some_ be deleted but others kept is also fine. I don't think that all the messages should be turned on.

dkfellows: Ugh; this is the compressor. Same principle though.

c_common/models/system_models/src/extra_monitor_support.c
dkfellows: There's `DEBUG_REINJECTOR`, `DEBUG_DATA_IN` and `DEBUG_DATA_OUT`.

c_common/models/compressors/src/bit_field_compressor.c
andrewgait: I'd be inclined to keep all of them but perhaps use the same flag everywhere.  Maybe even the same flag (#if LOG_LEVEL >= LOG_DEBUG) that's used in sPyNNaker, rather than a separate one for each.

PR #842, opened 2021-07-30
A client for talking to the Java Spalloc Server

dkfellows:
See SpiNNakerManchester/JavaSpiNNaker#325

It has just enough implemented to support FEC, no more; the higher-level support code isn't there yet, and the interface isn't complete at all.


dkfellows:
Example of usage of the basic code:

```py
client = SpallocClient("http://spinnaker.man/", "myself", "pAssw0rd")
print(client.version)
print(client.list_machines())
```

(I've had that much working… with a different service, username and password, of course)

dkfellows:
How to enable for normal usage in the config file:
```python
spalloc_server = http://myself:pAssw0rd@spinnaker.cs.man.ac.uk:8000/spalloc
```
Though the username (`myself`) and password (`pAssw0rd`) are in the URL there, they're parsed out by the client code. Communications will be secure if https is configured. _The new client is used if `spalloc_server` looks like an http or https URL, otherwise the old client is used._

`spalloc_port` and `spalloc_user` are ignored with the current version of the code; encoding the port in the URL seems more sensible (and the old code didn't use authentication at all, so there were no passwords).

dkfellows:
Added support for using the old-style service via URLs too:
```python
spalloc_server = spalloc://myself@spinnaker.cs.man.ac.uk:22244
```

dkfellows:
Includes websocket client for proxying. See SpiNNakerManchester/JavaSpiNNaker#444

dkfellows:
It occurred to me last night (and I'm writing it down here so I don't forget) that the low-level spalloc client for the new spalloc should be moved to SpiNNMan so that it can be used to construct a transceiver out of proxied connections.

rowleya (approved):
This is fine by me now, but I have made changes, so if anyone else wants to review, please do!

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -249,6 +247,27 @@ def __signal_handler(self, _signal, _frame):
...
+    @property
+    def __bearer_token(self):
+        """
+        :return: The OIDC bearer token
+        :rtype: str or None
+        """
+        # Try using Jupyter if we have the right variables
+        jupyter_token = os.getenv("JUPYTERHUB_API_TOKEN")
+        jupyter_ip = os.getenv("JUPYTERHUB_SERVICE_HOST")
+        jupyter_port = os.getenv("JUPYTERHUB_SERVICE_PORT")
```
andrewgait, line 259: Do we need to check all of these - surely one of them (the token?) will suffice?

andrewgait (approved):
This looks fine to me, just had one question, see comment...

spinn_front_end_common/interface/abstract_spinnaker_base.py
rowleya: I copied this from the HBP code - this should work on the EBRAINS jupyter.  The "jupyter_token" is used as authorization against the jupyter port and ip to get the real access token (see the code below).

dkfellows:
All the dependencies of this are now merged. Is this one ready to go too?

PR #843, opened 2021-08-16
Compressors sdram

rowleya:
Allows the simple compressor to use SDRAM.  This can be useful if the table is very big, but also includes lots of default entries, so it doesn't fit in DTCM, but will actually load directly into the routing tables without compression.

Also fixes the exiting mechanism to simply use spin1_exit but only if not doing an RTE.

Testing with: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/activity?branch=compressors_sdram


andrewgait (approved):
Looks good to me.

PR #844, opened 2021-08-17
Use the python-tools action in order to fix pip version

andrewgait:
The pip version currently needs to be fixed to an earlier version to avoid resolver issues (see https://github.com/SpiNNakerManchester/sPyNNaker/issues/1105).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,6 +43,8 @@ jobs:
       - name: Set dynamic environment variables
         run: |
           echo "SPINN_DIRS=$PWD/spinnaker_tools" >> $GITHUB_ENV
+      - name: Install pip, etc
+        uses: ./support/actions/python-tools
       - name: Checkout SpiNNaker Dependencies
         uses: ./support/actions/checkout-spinn-deps
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       - name: Set dynamic environment variables
         run: |
           echo "SPINN_DIRS=$PWD/spinnaker_tools" >> $GITHUB_ENV
+      - name: Install pip, etc
+        uses: ./support/actions/python-tools
```
dkfellows, line 47: Should also specify:
```
with:
    coverage: false
```
as we don't need support for handling coverage in this build.

dkfellows (changes requested):

.github/workflows/publish.yml
dkfellows: I got that by reading the source at https://github.com/SpiNNakerManchester/SupportScripts/blob/master/actions/python-tools/action.yml

dkfellows: Otherwise, looks good; the right sort of fix.

andrewgait: Done.

PR #845, opened 2021-08-19
Add app ID to database

rowleya:
Minor change to add the App ID to the external device database.  This allows external tools to send the sync signals required to progress a stepped simulation.  This will help to support the MUSIC integration with SpiNNaker.

Testing being done here:
SpiNNakerManchester/IntegrationTests#71


andrewgait (approved):
Looks fine to me.

PR #847, opened 2021-09-14
No executor

Christian-B:
This is the first version at this point mainly for review and comments.

Includes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/875

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1112
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/199
https://github.com/SpiNNakerManchester/MarkovChainMonteCarlo/pull/28
https://github.com/SpiNNakerManchester/SpiNNGym/pull/32
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/199

Must be done at or before:
https://github.com/SpiNNakerManchester/PACMAN/pull/409

Same branch name but probably independent of:
https://github.com/SpiNNakerManchester/SpiNNUtils/pull/162


Christian-B:
As much as possible I have retained the order the algorithms are run and the checks to say when they do or dont as close to master as possible.
The exception is in the getMachine part which has been simplified into a single run.

If the order or here is wrong and different to master it should be changed in this pr.
if the order is weird but like master it should be added to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/849

Same for the gates on when to run or not.

rowleya (commented):
I can see the idea here and it looks like it is on the right track.  In particular, the fact that the "main" methods of ASB are now just a linear call of functions certainly makes them easier to follow.  A couple of thoughts (which might be repeats of previous ideas):

1. Each of the calls to an algorithm" now also requires a function in ASB, which seems to make things quite complex and long.  To avoid this, it might be better that the algorithm itself checks whether it should be run rather than putting this outside of the algorithm.  Although this is more for the algorithm to do, as the user has to add the algorithm and the calling code anyway, it doesn't make as much sense to do this outside of the algorithm.  I would guess this might come with the global data changes being done elsewhere though, but I thought it should be mentioned just to make sure.  Note that where there is a selection between several algorithms with the same function, the function call is fine e.g. placer etc., but then a function call to each placer wouldn't be needed ideally.
1. The algorithms probably need to become functions rather than classes, as they are always instantiated and then called, which is a bit odd!  Making them functions would make removing the functions to call each algorithm easier.
1. FEC Timer might want to be renamed to something else, as it seems to be holding state here rather than just being a timer.  Maybe an FECState or something like that would make more sense.
1. I assume the whole of the XML file description can now go.

Christian-B:
"Each of the calls to an algorithm" now also requires a function in ASB, which seems to make things quite complex and long. To avoid this, it might be better that the algorithm itself checks whether it should be run rather than putting this outside of the algorithm. Although this is more for the algorithm to do, as the user has to add the algorithm and the calling code anyway, it doesn't make as much sense to do this outside of the algorithm. I would guess this might come with the global data changes being done elsewhere though, but I thought it should be mentioned just to make sure. Note that where there is a selection between several algorithms with the same function, the function call is fine e.g. placer etc., but then a function call to each placer wouldn't be needed ideally."

Reply:
I very much like the approach of making the algorithm decide if it should run and in many cases that is a better approach.  

It was not done yet to reduce the amount of change in one go,

Do we do it in a second pass after merging or before?

Christian-B:
"The algorithms probably need to become functions rather than classes, as they are always instantiated and then called, which is a bit odd! Making them functions would make removing the functions to call each algorithm easier."

Absolutely agree. Would suggest doing this at the same time as moving the should I run decision down.

Christian-B:
"FEC Timer might want to be renamed to something else, as it seems to be holding state here rather than just being a timer. Maybe an FECState or something like that would make more sense."

I am not at  all tied to the name.

But I point out the only "state" it holds is pointer back to _simulator and provenance_path only so that these do not need to be passed in every time.  Gloabal Data model may remove the need for that.

_start_time is needed for a timer

_algorithm and _category are purely for logging.

rowleya:
> Do we do it in a second pass after merging or before?

That is the question as always!  I am not against merging this without this change.  I think that depends how many different things you want to depend on these changes.  If you have multiple things going forward, it would likely be easier for the changes to be in master first.

Christian-B:
"I assume the whole of the XML file description can now go."

Absolutely.

I did not remove the xml readers because well I forgot.

I did not remove the xml files themselves for now as they where very handy references to how it worked in the past.

rowleya:
> "FEC Timer might want to be renamed to something else, as it seems to be holding state here rather than just being a timer. Maybe an FECState or something like that would make more sense."
> 
> I am not at all tied to the name.
> 
> But I point out the only "state" it holds is pointer back to _simulator and provenance_path only so that these do not need to be passed in every time. Gloabal Data model may remove the need for that.
> 
> _start_time is needed for a timer
> 
> _algorithm and _category are purely for logging.

I think it was the "skip" checks that are in here that make it a bit odd in a timer.  I can see that these don't actually do anything based on internal state now though that you have pointed this out.

Christian-B:
One idea that is python but ugly and I have confirmed works is to have algorithms that decide to skip to raise an AlgorithmSkippedException.  the FecTimer __exit__ would then catch this by checking the type param.

The FecTimer would then do the logging/ provenance required and by returning True the with statement does not reraise the exception.  Any other Exception would also be logged/ provenance as required but by __exit__ returning False the with reraises the Exception.

As toy example to show how this works is:

class ExceptionEater(object):

    def __enter__(self):
        return self

    def __exit__(self, type, value, traceback):
        return type == ZeroDivisionError
        #type is not None


with ExceptionEater():
    pop = 1/0
with ExceptionEater():
    a + b


the above is run exits with a "NameError: name 'a' is not defined"

rowleya:
I don't think this is too ugly and not much different from calling a skip method during the run.  It has the advantage that the skip exception then stops the flow as well rather than having "if x_skip" calls which look a bit odd.

rowleya:
As an example:
```
if timer.skip_if_value_already_set(self._machine, "machine"):
    return
if timer.skip_if_value_false(self._use_virtual_board, "use_virtual_board"):
    return
```

would become:
```
if self._machine is not None:
    raise AlgorithmSkipException("Machine already exists")
if not self._use_virtual_board:
    raise AlgorithmSkipException("This algorithm is only needed for virtual boards")
```

Christian-B:
The real advantage of using raise AlgorithmSkip that the algorithm can raise the AlgorithmSkip without having access to the FecTimer

Then the FecTimer can take care of whatever reporting/logging is desirable.

PR #865, opened 2021-09-20
avoid update_buffer being called until do run

Christian-B:
ReverseIPTagMulticastSourceMachineVertex._fill_send_buffer is not needed until do_run

So to avoid calling it early this PR.
1. Changes ReverseIPTagMulticastSourceMachineVertex.update_buffer to be triggered by the injection of RunUntilTimeSteps
2. Removed _fill_send_buffer from ReverseIPTagMulticastSourceMachineVertex.generate_data_specification
3. No longer sets RunUntilTimeSteps in the do_data_generation stage.

This has the added advantage that
RunUntilTimeSteps is now always the end of the do_loop timestep

In the past update_buffer did not fill the send buffer if there was no virtual_key
However if there is no virtual key
1. The user has probably done something weird as why have a ReverseIPTagMulticastSourceMachineVertex with no output
2. The user may be recording the spikes and then the buffers still need filling

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -1674,7 +1674,6 @@ class AbstractSpinnakerBase(ConfigHandler):
         # The initial inputs are the mapping outputs
         inputs = dict(self._mapping_outputs)
         tokens = list(self._mapping_tokens)
-        inputs["RunUntilTimeSteps"] = n_machine_time_steps
         inputs["FirstMachineTimeStep"] = self._current_run_timesteps
 
         # This is done twice to make things nicer for things which don't have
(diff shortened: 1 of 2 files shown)


Christian-B:
In master this script fails showing the approach was not just wasteful but wrong.

import spynnaker8 as sim
sim.setup(timestep=1.0)
input = sim.Population(1, sim.SpikeSourceArray(spike_times=[1, 12, 23]), label="input")
input.record("spikes")
sim.run(10)
sim.run(10)

Although the error message was weird:
The buffer size 61489290 is too small for any data to be added for region 3 of vertex input:0:0

In this pr with update_buffer running even with no virtual_key that works

Christian-B:
Better fix in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/882

Confirmed the above example works in that pr

PR #870, opened 2021-09-23
Spalloc avoid boards

Christian-B:
Adds the ability for a user to specif which boards to avoid as the host/ chip 0,0 in spalloc.

Does not include HBP at this point.

Known risk is that if the combination of job size and avoid list is too big it could block the whole machine!

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
+++ b/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
@@ -145,7 +145,7 @@ class SpallocAllocator(object):
         if spalloc_machine is not None:
             spalloc_kw_args['machine'] = spalloc_machine
 
-        job, hostname = self._launch_job(n_boards, spalloc_kw_args)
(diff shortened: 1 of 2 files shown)


rowleya (approved):
Fine when Integration tests pass

spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
```diff
@@ -148,18 +153,32 @@ def __call__(
             False, None, None, machine_allocation_controller
         )
 
+    def _launch_checked_job(self, n_boards, spalloc_kw_args):
```
rowleya, line 156: I don't think this gets called?

Christian-B: oops

Christian-B: fixed and manually tested now

PR #883, opened 2021-10-26
Always have an Application graph

Christian-B:
This PR makes sure there is always an application Graph.

GFE calls to add machine Vertices and Edges are wrapped in Application ones.

Yes this then has the weirdness that these have to be partitioned,
but allows algorithms to always use the Application Graph.

This PR is only ground work as it guarantees the presence of an Application Graph but does not yet make use that fact.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/82


Christian-B:
This is now part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/884 which if this goes in should too.

So closing this part pr

PR #884, opened 2021-10-27
No need for code in case there is no app graph

Christian-B:
This PR guarantees there is always an Application graph.

But at the price that GFE users will see (fast) partitioning even when they only supplied machinve vertices.

This PR alone works but is in my opinion not enough of a justification by itself.
But does for a base for any other branch that NEEDS application graphs.

Best done at the same time as:
https://github.com/SpiNNakerManchester/PACMAN/pull/411
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1122

Removes the placer_reports_without_application_graph
as they have the same data as the ones with application_groah that can now always bet written


Christian-B:
done elsewhere

PR #885, opened 2021-10-27
Add provenance writing for app_vertex

andrewgait:
Part of https://github.com/SpiNNakerManchester/sPyNNaker/pull/852


Christian-B:
Should the filed and param "population" not be "label".

At the FEC level there are no Populations (minor)

You are actually saving the AbstractPopulationVertex.label.  (Yes I know that is typically also the population label)

Would allow the table/ method also to be used for none Population Vertices

PR #886, opened 2021-10-28
Global Data model

Christian-B:
At this point this is an example of how the global data object would look.

Model
- One (or more DataModel) singleton classes with only double underscore variables

View
Wrapping the model(s) with methods that raise the correct exception if the data is not available

DataStatus
What state the data is in and throwing the correct Exception

Writer
Extends View but also adds methods to change the state and data
- this would included validation that data is legal ect.

--

There will be more that one model for example FecDataModel, SpyDataModel
See https://github.com/SpiNNakerManchester/sPyNNaker/pull/1125
They would be independent each with their own data
A model could also be a database

There will be more than one Viewer where the later ones will extend the earlier ones therefor exposing both sets of methods in one API
For example 
class SpyDataView(FecDataView):
which adds methods for the SpyDataModel

There will be more than one Writer where the later extend both the Viewer at its level and the Writer below.
For example
class SpyDataSetter(SpyDataView, FecDataSetter):
which adds methods for changing the SpyDataModel

Technically the read (@property) methods could easily be classless just be imported where needed.
Even the write methods could be classes.
Much in the same style as the config_holder ones 

However I definitely don't want the write methods to be classless as I want to discourage use of these outside of the sim and ASB stacks. I prefer having clear View Objects we can pass to users and algorithms to highlight this is a bound together set of data and not stuff from all over the place.

### DIFF
--- a/spinn_front_end_common/data/__init__.py
+++ b/spinn_front_end_common/data/__init__.py
@@ -17,4 +17,4 @@ from .fec_data_view import FecDataView
 from .fec_data_writer import FecDataWriter
 
 
-__all__ = ["FecDataView"]
+__all__ = ["FecDataView", "FecDataWriter"]
--- a/spinn_front_end_common/data/__init__.py
+++ b/spinn_front_end_common/data/__init__.py
@@ -1,0 +1,20 @@
+# Copyright (c) 2017-2019 The University of Manchester
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


Christian-B:
When compared to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/887 @rowleya says "I think the new one makes sense".

As I agree with him I am closing this PR.

PR #887, opened 2021-11-02
Global Data model Version2

Christian-B:
Alternative to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/886
See also https://github.com/SpiNNakerManchester/sPyNNaker/pull/1126

Advantages:
- No need for the self._fec_data._Fec_Data_foo references

Disadvantages
View and Writer must have a different __fec_data variable
_Fec_Data class in no longer protected by a double underscore


Christian-B:
Do we prefer 1. FecDataView.foo  2. FecDataView.get_foo()  or 3 both where the property if no value is available raises an Exception while the get returns None.    If we don't go for Option 3 then I will keep/add FecDataView.has_foo()

Christian-B:
now part of much bigger pr

PR #892, opened 2021-11-15
put recover_from_error back in

Christian-B:
Please test this on the branch with errors in it

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -3112,8 +3112,8 @@ class AbstractSpinnakerBase(ConfigHandler):
                 except Exception as stop_e:
                     logger.exception(f"Error {stop_e} when attempting to stop")
 
-                # reraise exception
-                raise run_e
+            # reraise exception
+            raise run_e
 
     def _recover_from_error(self, exception):
         """
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -3101,8 +3101,7 @@ class AbstractSpinnakerBase(ConfigHandler):
             self._shutdown()
             sys.exit(1)
         except Exception as run_e:
-            e_inf = sys.exc_info()
-            self._recover_from_error(run_e, e_inf)
+            self._recover_from_error(run_e)
 
             # if in debug mode, do not shut down machine
             if get_config_str("Mode", "mode") != "Debug":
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -3602,10 +3642,12 @@ def stop(self, turn_off_machine=None,  # pylint: disable=arguments-differ
...
-            self._do_stop_workflow()
-            # TODO recover from errors?
-            # self._recover_from_error(
-            # e, exc_info[2], executor.get_item("ExecutableTargets"))
+            try:
+                self._do_stop_workflow()
+            except Exception as e:
+                exn = e
+                exc_info = sys.exc_info()
+                self._recover_from_error(e, exc_info[2])
```
rowleya, line 3650: This is different from the other call to this function; I think it is enough to just pass the exc_info.

spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -3177,12 +3217,12 @@ def _recover_from_error(self, exception, exc_info, executable_targets):
...
-                logger.exception("Could not read provenance")
+            except Exception as pro_e:
+                logger.exception(f"Could not read provenance due to {pro_e}")
 
         # Read IOBUF where possible (that should be everywhere)
         iobuf = IOBufExtractor(
-            self._txrx, executable_targets, self._executable_finder)
+            self._txrx, self._executable_targets, self._executable_finder)
         try:
             errors, warnings = iobuf.extract_iobuf()
```
rowleya, line 3227: I think the "print" call should also go in here, otherwise it will fail because things are not defined.

rowleya: Ah no, I see now that it does define them in the except.  Odd, but OK...

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: yes I saw that too.  Release 6.0.0 had it that way

rowleya: I'm not sure how critical it is, but it should probably be consistent.  This seems to just go in to logger.error, and the documentation of that seems to say it should just be sys.exec_info()...

rowleya:
This works on my test branch, which fails due to an RTE (a forced one from the init stage).

Christian-B:
is it work expanding at least one of gfe_integration_tests/test_rte to check provenance files are written.

I checked manually with and without the recover call and it does change.

rowleya (approved):
Happy for this to go in if you are happy it has been tested enough

Christian-B:
I am going to merge this with one other PR an run jenkins

Christian-B:
merging as approved and passing Jenkins

PR #893, opened 2021-11-15
App graph is never None but may be empty

Christian-B:
In the release 6.0.0 if there was no app vertices the algorithms would be passed a None

No app vertices can come from either GFE machine level only or an empty graph.

The no executor branch changed the behaviour to always pass the app graph even if empty.

This made the algorithms that just for loop the app graph a little cleaner.

The issue was the algorithms that have an app and a machine version.

As there is separate work which may remove the machine_graph or at least require there always to be an app graph the choice is to change the if app_grap is None test ti ones that check n_vertices.

If this PR is refused ASB must revert to checking app_graph n_vertices before passing it.

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/insert_edges_to_extra_monitor_functionality.py
+++ b/spinn_front_end_common/interface/interface_functions/insert_edges_to_extra_monitor_functionality.py
@@ -65,7 +65,7 @@ class InsertEdgesToExtraMonitorFunctionality(object):
             "Inserting edges between vertices which require FR speed up "
             "functionality.")
 
-        if application_graph == 0:
+        if application_graph.n_vertices == 0:
             for vertex in progress.over(machine_graph.vertices):
                 if isinstance(vertex, ExtraMonitorSupportMachineVertex):
                     self._process_mach_graph_vertex(vertex, machine_graph)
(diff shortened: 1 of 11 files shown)


spinn_front_end_common/interface/interface_functions/insert_edges_to_extra_monitor_functionality.py
```diff
@@ -67,7 +65,7 @@ def __call__(self, machine_graph, placements, machine,
             "Inserting edges between vertices which require FR speed up "
             "functionality.")
 
-        if application_graph is None:
+        if application_graph == 0:
```
rowleya, line 68: Probably want to check the property here

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
```diff
@@ -65,7 +65,7 @@ def __call__(
                 progress, machine, machine_graph, vertex_to_chip_map)
 
         # progress data receiver for data extraction functionality
-        if application_graph is not None:
+        if application_graph.n_verticese > 0:
```
rowleya, line 68: Mis-spelled

rowleya (changes requested):

spinn_front_end_common/interface/interface_functions/insert_edges_to_extra_monitor_functionality.py
Christian-B: good spot. test got it too Fixed.

spinn_front_end_common/interface/interface_functions/insert_extra_monitor_vertices_to_graphs.py
Christian-B: Yep. Tests got that too. Thanks.

Christian-B:
Will merge with another pr and run jenkins

rowleya (approved):

PR #901, opened 2021-12-01
Key translation

rowleya:
Updates Live Packet Gather to add a translation function.  This allows multiple keys coming from different cores to be translated into simple neuron IDs.  Other features already present in the LPG then allow a global key to be added to this if required.

See SpiNNakerManchester/sPyNNaker#1138 for integration test, which also shows the key prefixing (note this is not required, just an option).


andrewgait (approved):
This all looks fine too; as stated on the sPyNNaker branch, running the integration tests as a test is being added is probably a good idea...

PR #909, opened 2022-01-18
Dont save config values in ASB

Christian-B:
This pr stops the saving of 
get_config_str("Machine", "machine_name")
get_config_str("Machine", "spalloc_server")
get_config_str("Machine", "remote_spinnaker_url")

as Values in AbstractSpinnakerBase

Instead the values are retrieved from configs as needed.

While this is similar to the global data work it is in a separate PR due to a few minor code changes.

hostname param has been removed.
Due to safety checks it would rarely have worked even had it been used.

repr methods just use ipaddress as if there is a machine_name (ex self._hostname) ipaddress is set to that anyway

set_up_machine_specifics replaced with check_machine_specifics 
Checks and errors should sill be the same.
Not moved inside ASB init as it depends on reading the spynnaker.cfg or spiNNakerGraphFrontEnd.cfg file 

default n_boards for spalloc removed from GFE
calling get_machine before run will now fails if on spalloc and no n_boards/n_chips provided. Same as Spynnaker

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1145
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/207

Tested by
https://github.com/SpiNNakerManchester/IntegrationTests/pull/94

### DIFF
--- a/fec_integration_tests/file_convertor_tests/test_write_json_machine.py
+++ b/fec_integration_tests/file_convertor_tests/test_write_json_machine.py
@@ -143,7 +143,6 @@ class TestWriteJson(unittest.TestCase):
         set_config("Machine", "spalloc_server", self.spalloc)
         set_config("Machine", "spalloc_port", self.spin2Port)
 
-
         try:
             (hostname, version, _, _, _, _, _, m_allocation_controller) = \
                 spalloc_allocator(n_chips=20)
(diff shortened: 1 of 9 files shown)


Christian-B:
As described above. hostname completly removed as is default n-boards on GFE

Issue #866, opened 2021-09-21
Energy Report vs multiple _do_run


Christian-B:
The energy report is only really designed for a single _do_run clcyle

The three options are.

1. Only run the Energy report stuff on the first _do_run cycle
2. Nuke the whole energy report as noone uses it
3. Fix the Energy Report

andrewgait:
I’d much rather keep this if we can - and if people aren’t using it then we should be directing them to do so.

Christian-B:
Yes I do see the point in keeping the Energy Report.

Is it worth upgrading it to mutliple _do_run cycles or is the information from a single do_run enough?

andrewgait:
I think it's worth upgrading to multiple do_run cycles, though whether you can distinguish at this level between multiple calls to run() and auto-pause and resume ought to be considered.

Christian-B:
In master the report runs every do_run cycle regardless if it is and auto pause or mutliple sim.run.

Master also does not reset timings on a hard reset. It should.

The time execute_time depends on when the executor decides to run FinaliseTimingData

Christian-B:
With the no executor stuff it is worth reconsidering the phases reported.

Christian-B:
What stages o report is also related to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/849

Christian-B:
Care is need to make sure _extraction_time and run_time do not both include buffer extraction.

Christian-B:
Also looks like the EnergyReport goes boom on runtime is None

So either fix that or add a condition to run the report that runtime is not None

Christian-B:
With the removal of the pacman executor and _pacman_provenance the engery report is seriously broken so currently (in no executor work) deactivated.

As provenance is being changed in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/867 it is not worth fixing the "CHEAT!" in ComputeEnergyUsed until then,

Christian-B:
Turn back on spynnaker_integration_tests/test_power_monitoring

Christian-B:
the energy report also completely ignores the get machine time

Christian-B:
I also notice that buffer_manager adds extraction_timing even when it is inside the run loop.

That may be due to BufferExtractor being move during the removal of the pacman exectuor

PR #912, opened 2022-01-21
Category timing into provenance database

Christian-B:
Warning Energy reporting was broken and still is https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/866
Just a little less.

This PR replaced the ASB._xxx_time variables with data in the Provenance databse

ASB and BufferManager write the data in.

Energy function read it out ASAP.  this could often be done later but cleaning up the Energy Stuff is out of scope.

One change it that timing from before a hard reset which where not cleared now are.


Christian-B:
tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/category_timing/1/pipeline

PR #918, opened 2022-02-11
Add tags where malloc is used

rowleya:
Add tags to SDRAM allocations to make use of SpiNNakerManchester/spinnaker_tools#159.  This ensures any allocations that are done will survive retries of the SCP messages.


dkfellows (approved):
Ready when `spinnaker_tools` is ready.

PR #932, opened 2022-03-22
Send log messages to the database

Christian-B:
send all log messages to the provenance databse.

https://github.com/SpiNNakerManchester/SpiNNUtils/pull/174

This includes safety code if the database is locked.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/137


spinn_front_end_common/interface/provenance/db.sql
```diff
@@ -218,3 +218,31 @@ CREATE TABLE IF NOT EXISTS boards_provenance(
...
+VALUES
+    (50, "CRITICAL"),
+    (40, "ERROR"),
+    (30, "WARNING"),
+    (20, "INFO"),
+    (10, "DEBUG");
+
+CREATE VIEW IF NOT EXISTS p_log_view AS
+    SELECT
+		name,
```
dkfellows, line 244: Use tabs or don't use tabs, but don't mix.

spinn_front_end_common/interface/provenance/db.sql
```diff
@@ -218,3 +218,31 @@ CREATE TABLE IF NOT EXISTS boards_provenance(
     ip_addres STRING NOT NULL,
     ethernet_x INTEGER NOT NULL,
     ethernet_y INTEGER NOT NULL);
+
+---------------------------------------------------------------------
+-- A table to store log.info
+CREATE TABLE IF NOT EXISTS p_log_provenance(
```
dkfellows, line 224: Should this table include the timestamp as its own column?

Christian-B: good question I dont know

spinn_front_end_common/interface/provenance/db.sql
Christian-B: fixed

Christian-B:
> Is what is being logged controlled by the general logging level? I ask because logging stuff at `DEBUG` level is likely to produce a lot of write activity (and may have quite a profound effect on performance).

Currently yes it does . because I had not considered that.

An issue is that in FormatAdapter.log has self.isEnabledFor(level) even for level 10 (debug)

Christian-B:
If https://github.com/SpiNNakerManchester/SpiNNUtils/pull/177 is approved it will be easy to only write to database if the message is of interest.

spinn_front_end_common/interface/provenance/db.sql
Christian-B: Done

rowleya (approved):

PR #939, opened 2022-04-20
We do support runforever more than once

Christian-B:
Alternative to https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/213

Turns out do even test multiple run forever.

As there Not Implemented was more of a check of do we need to worry it.

As it worked before no good reason to remove the functionality,

tested by 
https://github.com/SpiNNakerManchester/IntegrationTests/pull/116


andrewgait:
It looks like PDP2 also has multiple run_until_complete examples, so definitely worth doing this way.

PR #940, opened 2022-05-10
Always app graph

Christian-B:
Cherry pick from https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/884/

This PR guarantees there is always an Application graph.

But at the price that GFE users will see (fast) partitioning even when they only supplied machine vertices.

The main purpose of this PR is to support the work in reduce overhead.

That is why all the check for Machine vs application graph are left in to avoid clashes.

Tested by
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/always_app/1/pipeline


Christian-B:
done better elsewhere

PR #947, opened 2022-06-06
Physical board chip report

rowleya:
This adds the physical core of the monitor to the board-chip report, which can help in blacklisting when a monitor crashes during an operation.

### DIFF
--- a/spinn_front_end_common/utilities/report_functions/board_chip_report.py
+++ b/spinn_front_end_common/utilities/report_functions/board_chip_report.py
@@ -55,4 +55,4 @@ def _write_report(writer, machine, progress_bar):
 
         writer.write(
             "board with IP address : {} : has chips [{}]\n".format(
-                e_chip.ip_address, xyps.join(", ")))
+                e_chip.ip_address, ", ".join(xyps)))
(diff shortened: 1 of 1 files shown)


spinn_front_end_common/utilities/report_functions/board_chip_report.py
```diff
@@ -49,8 +49,10 @@ def _write_report(writer, machine, progress_bar):
     :param ~spinn_machine.Machine machine:
     :param ~spinn_utilities.progress_bar.ProgressBar progress_bar:
     """
-    for chip in progress_bar.over(machine.ethernet_connected_chips):
-        xys = machine.get_existing_xys_on_board(chip)
+    for e_chip in progress_bar.over(machine.ethernet_connected_chips):
+        xyps = [f"({chip.x}, {chip.y}, P: {chip.get_physical_core_id(0)})"
```
Christian-B, line 53: Why the hardcode P 0?

rowleya: It is virtual processor 0 that I want to record the ID of, as that is the monitor.  If any other core crashes, we can likely ask the machine for the details, but if the monitor crashes, this will be the only record.

PR #958, opened 2022-07-21
Stop energy report crashing when there's no data extraction

andrewgait:
Energy reporting was crashing in cases where no data extraction was taking place (e.g. where no variables were set to record).  This PR stops the crash but doesn't fix what we believe are underlying problems with the energy reporting as it currently stands; some comments and notes have been added in the reporting code here along those lines, and we will likely revisit how the timing in particular is carried out in future work.


rowleya (approved):
Looks OK

PR #960, opened 2022-07-26
Remove Max Machine

Christian-B:
The only partitioner now no longer needs a Machine or ResourceContainer (which needed a Machine)

The Placer now runs after we get a real machine

Therefor there is never a need for a max machine

tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/no_max_machine/2/pipeline/


andrewgait (approved):
Looks good to me, happy to merge now.

Issue #954, opened 2022-07-11
Review adding to placements


Christian-B:
In https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/950 a lot of functions are now adding placements instead of things to a graph.

Once https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/929 is in we shoudl review what is the best way to handle this adding of placements.

Christian-B:
insert_extra_monitor_vertices_to_graphs will be the most interesting as it adds placements AND returns two dicts

Christian-B:
Thinking about this I am now considering.
1. keeping the passing of a Placements object into the algorithms that add Placement objects
2. Grouping get_machine and Placements creating Algorithms into a separate ASB stages.
     Then allowing this stage to loop back and get a bigger Machine if needed.

Christian-B:
With better models for computing how large a machine in needed this is no longer required.

PR #969, opened 2022-08-09
used a fixed machine where possible

Christian-B:
This PR. Changes how get_machine works.

It now treats fixed size machine different from flexible size machine.

A fixed size is any of:
- A virtual machine (they always have size)
- Boards directly connected (cfg machineName value set)
- Where setup is provided with either a n_boards_required or n_chips_required

When you have a fixed size
- Create and save the Machine during setup.
- It is then never changed until a new setup.
- Don't record is user has access the machine as it doe not matter
- Machine has not affect on reset and reset has not affect on machine

When you have a flexible machine
- Create machine after partitioning(In the future maybe even more than once (see https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/954)
- record if the user has seen the machine
- A soft reset is not allowed if user has seen machine. ASB make is hard
- Raise exception if user tires to access machine between reset and run
- Drop the existing machine on a hard reset

This has the benefit that a spalloced machine obtained with n required will not change and remain available even with a hard reset.

This  drops the generator which allowed the View to call ASB

The **disadvantages** is that now a number (currently 69) of the sPyNNaker unittests will attempt to get a Machine
- The ones that do sim.setup but not sim.run will now try to obtain a machine.
- If the cfg points to spalloc there is a flexible machine and a machine is not obtained.
- If the cfg points to a machineName all these unitests slow down as they obtain a machine, and even fail if that machine is not connected.
- If the cfg is invalid as https://github.com/SpiNNakerManchester/SupportScripts/commits/master/actions/configure-spynnaker was the unittests will now fail.  
 

https://github.com/SpiNNakerManchester/SpiNNMachine/pull/178
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1211
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/222

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/128


Christian-B:
https://github.com/SpiNNakerManchester/SupportScripts/commits/master/actions/configure-spynnaker fixed direct to master as impossible to test branches on that repository.

Christian-B:
This has blown up more than I had hoped.

The transceiver is then also kept around.

What to change and not change on the boards is prone to errors.

This starts to add a third reset type and is not worth it!

Christian-B:
got too messy

PR #977, opened 2022-09-21
Fix power monitor buffer extraction

Christian-B:
The data extraction from ChipPowerMonitorMachineVertex had been disabled due to a bug.

This PR and the associated ....

1. Add a new iterate_placements_by_vertex_type method
2. rename the previous one to iterate_placements_by_xy_and_type
    (Both on Placements and the View)
3. Use the new method where there either there is no Progressbar or where the ProgressBar did not make sense.
4. Merge buffer_extractor into BufferManager.
5. If using Java just use  iterate_placements_by_vertex_type  and avoid pre counting and storing
6. If using python count and store until https://github.com/SpiNNakerManchester/PACMAN/issues/468 decides to keep or remove
7. pre counter now uses iterate_placements_by_vertex_type as old version via AppVertex did not work as ChipPowerMonitorMachineVertex do not have app vertexes.

https://github.com/SpiNNakerManchester/PACMAN/pull/469

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/142


andrewgait (approved):
Looks good to me.

PR #984, opened 2022-10-03
Replace Multiple interwoven timers with more use of FecTimer

Christian-B:
Previously there where several Timers which where used for the Energy reports including the ones used for time data.

The problem was there where gaps between the time and the Data Generation was inside the Loading.

This PR instead adds more into the FecTimer as @classmethod  (so there is only 1 reachable anywhere)

There is now a category Timer which starts during Setup and runs until end.

Each time a category is started a row is added to the Provenance Data.
That includes a record of what run number and if applicable auto pause loop is currently active.
As well as if the Machine is On or not.
Each time it ends that row is updated with how long it took.

These Categories are nested so that at the end of a Category it will automatically return to the previous one.
For Example it goes from 
WAITING to RUN_OTHER to , MAPPING to GET_MACHINE back to MAPPING back to RUN_OTHER back (via other stuff) to WAITING 
This allows get_machine to start and end GET_MACHINE without having to worry if it should go back to MAPPING or WAITING (on a sim.get_machine() call.
It also allows Jumping into SHUTTING_DOWN on an Exception.
An Exception is raised if start_category and end_category calls do not match.
There is a final stop_category_timing which closes the last category and clears all starts with no matching end.
Note: How the EnergyReports use this data has been left as close to previous as possible (Must be cleaned up in later work)

The with FecTimer(...) calls have been changed so they
1. Automatically link to the Category and therefor machine_on, run, auto pause loop they are in.
2. Added a "work" param to better track what the Algorithm is actually doing.
For example this tracks which Algorithm are actually Loading data onto the machine rather than just reports in the do_load method.

FecTimer moved to the Provenance directory as that is what it is and for easier avoidance of import loops

ProvenanceReader and ProvenanceWriter (and db.sgl) changed to reflect this 

ASB self._n_loops  replaced with View run_step methods.
ASB.check_machine_specifics was unused so just a cleanup.

TODO: Either in this PR or more likely followup PRs.
Fix Energy Reports. Considering what actually need now that machine is never turned off mid run.
Add timer report.
Move timer/ energy reports outside of _do_run
Review Categories: Which do we actually care about
Review "work" names; Which do we actually care about
Add machine on/ machine off ProvenanceReader methods.

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1227

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/149


spinn_front_end_common/interface/buffer_management/buffer_manager.py
```diff
@@ -544,24 +542,22 @@ def __count_regions(self):
...
-                        self.__old_get_data_for_placements_with_monitors(
-                            recording_placements, progress)
-                    else:
-                        self.__old_get_data_for_placements(
-                            recording_placements, progress)
-        with ProvenanceWriter() as db:
-            db.insert_category_timing(BUFFER, timer.measured_interval, None)
+                    self.__old_get_data_for_placements(
+                        recording_placements, progress)
+        FecTimer.end_category(FecTimer.EXTRACTING)
```
rowleya, line 560: Should the start and end category match here?

spinn_front_end_common/interface/interface_functions/compute_energy_used.py
```diff
@@ -72,13 +72,13 @@ def compute_energy_used(machine_allocation_controller=None):
...
-    dsg_time = db.get_category_timer_sum(DATA_GENERATION)
-    execute_time = db.get_category_timer_sum(RUN_LOOP)
+    dsg_time = db.get_category_timer_sum(FecTimer.DATA_GENERATION)
+    execute_time = db.get_category_timer_sum(FecTimer.RUN_LOOP)
     # NOTE: this extraction time is part of the execution time; it does not
     #       refer to the time taken in e.g. pop.get_data() or projection.get()
-    extraction_time = db.get_category_timer_sum(BUFFER)
-    load_time = db.get_category_timer_sum(LOADING)
-    mapping_time = db.get_category_timer_sum(MAPPING)
+    extraction_time = db.get_category_timer_sum(FecTimer.BUFFER)
```
rowleya, line 79: There seem to be two categories for this - EXTRACTING and BUFFER.  Are you sure this is the right one?

spinn_front_end_common/interface/provenance/fec_timer.py
```diff
@@ -65,51 +69,85 @@ class FecTimer(object):
...
+    DATA_GENERATION = "data_generation"
+    MAPPING = "Mapping Stage"
+    RUN_LOOP = "Running Stage"
+    BUFFER = "BufferExtraction"
+    RESETTING = "Resetting"
+    SHUTTING_DOWN = "Shutting down"
+    ERROR = "Error"
+
+    CATEGORIES = [WAITING, SETTING_UP, RUN_OTHER, GET_MACHINE, MAPPING,
+                  LOADING, RUN_LOOP, BUFFER, RESETTING, SHUTTING_DOWN, ERROR]
```
rowleya, line 91: If these are different from the Work Type constants, it might be better to make an enum (I know it isn't enumerated, but would separate the types).  This might then be easier to identify incorrect use in the various points like start category vs. timer type.

spinn_front_end_common/interface/provenance/fec_timer.py
```diff
@@ -65,51 +69,85 @@ class FecTimer(object):
...
+    COMPRESSING = "Compressing"
+    CONTROL = "Control"
+    SYNAPSE = "Expanding Synapse"
+    RUNNING = "Running"
+    EXTRACTING = "Extracting"
+    REPORT = "Reporting"
+
+    WORK_TYPES = [
+        OTHER, GET_MACHINE, LOADING, BITFIELD, CONTROL, SYNAPSE, RUNNING,
+        EXTRACTING, REPORT]
```
rowleya, line 107: See above - might be better as an Enum

spinn_front_end_common/interface/provenance/fec_timer.py
```diff
@@ -172,23 +209,79 @@ def _stop_timer(self):
...
+        """
+        if cls._category_name != category_name:
+            if cls._category_name != cls.ERROR:
+                logger.exception(f"Incorrect end of category: {category_name}"
+                                 f" when in category{cls._category_name}")
+            cls._previous = []
+            cls._change_category(cls.ERROR)
+        else:
+            try:
+                previous = cls._previous.pop()
```
rowleya, line 280: Doesn't this mean that you need to have at least one category which you never end?  If so, would it make sense for that to go in the constructor as an option, or else in the _previous list by default?  It could then error if you try to take that out at any time.

unittests/utilities/test_fec_timer.py
```diff
@@ -58,6 +66,33 @@ def test_error(self):
...
+            FecTimer.MAPPING)
+        self.assertGreater(on, 0)
+        self.assertGreater(off, 0)
+
+    def test_mess(self):
+        FecTimer.end_category(FecTimer.WAITING)
+        FecTimer.start_category(FecTimer.RUN_OTHER)
+        FecTimer.start_category(FecTimer.MAPPING)
+        FecTimer.end_category(FecTimer.RUN_OTHER)
+        FecTimer.end_category(FecTimer.MAPPING)
```
rowleya, line 94: Shouldn't there be an error here?

rowleya (changes requested):
A couple of thoughts submitted.  It might also be nice to understand the difference between starting a category, and a timer for an algorithm doing work.  I *think* it means you can time a whole section vs. an algorithm, but then I can't see this being extracted from the database (I don't think anyway).

Christian-B:
> A couple of thoughts submitted. It might also be nice to understand the difference between starting a category, and a timer for an algorithm doing work. I _think_ it means you can time a whole section vs. an algorithm, but then I can't see this being extracted from the database (I don't think anyway).

Yes starting a Category is for a section. Which can have one or many Algorithms in it.
Also the Category timing does not allow gaps. There is always a category active even if just WAITING for the next user input.

Category timings are used by the energy_reports. Not at all as well as I would like)

This includes the summary_energy_report.rpt which after work in a alter pr will be a good report for recording how long each section took.

spinn_front_end_common/interface/interface_functions/compute_energy_used.py
Christian-B: Agreed.

Proves that the timer does not die even on bad info.
Does Log
2022-10-04 14:35:54 ERROR: Incorrect end of category: Extracting when in categoryBufferExtraction

But does need fixing.

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: I had debated enums.
Will change.

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: yep

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: There should be a WAITING started during sim.setup.
(SHOULD! oops)
Before setup time is of zero interest.

It does error if you get out the last category but TOO silently.
Should at the very least log.exception

unittests/utilities/test_fec_timer.py
Christian-B: it does Error but does it with just a log and changing to ERROR.

Discussed with @rowleya and decided to actually Error!

spinn_front_end_common/interface/buffer_management/buffer_manager.py
Christian-B: Yes start and end should match.

Note: In this case I removed the Category BUFFER/EXTRACTING so this is now part of the RUN_LOOP
Instead added a Work type  EXTRACT_DATA  and used that in the EnergyReport  (for now)

spinn_front_end_common/interface/interface_functions/compute_energy_used.py
Christian-B: as above this mismatch is completely gone.
But if it happened again the FecTimer goes boom!

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: Created TimerCaterory and TimerWork enums.

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: Created TimerCaterory and TimerWork enums

spinn_front_end_common/interface/provenance/fec_timer.py
Christian-B: There now is one category WAITING which starts during setup which is end_category is called will raise an Exception.

There is now also a stop_category_timing which cleanly closes down the running Category and returns the FecTimer into the state it was before setup was called. 
That works no matter how many start_category calls are waiting for  end_category

PR #985, opened 2022-10-04
Checksum fixes

rowleya:
Add checksums to the region reloading code, and move the verification of checksums to somewhere that is used on reload.  Note that this resets the checksum to 0 after checking.  This is useful for regions which are changed during run, and avoids issues if the binary is reloaded later.

Depends of SpiNNakerManchester/DataSpecification#124


spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
```diff
@@ -29,96 +28,149 @@
...
-
-        # build file paths for reloaded stuff
-        run_dir_path = FecDataView.get_run_dir_path()
-        self._data_dir = generate_unique_folder_name(
-            run_dir_path, "reloaded_data_regions", "")
-        if not os.path.exists(self._data_dir):
-            os.makedirs(self._data_dir)
-
-        report_dir = None
-        if get_config_bool("Reports", "write_text_specs"):
```
Christian-B, line 61: If you get rid of write_text_specs here you should also do it in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/c004357a308bf2f0514aa7580fe148b07ae349e0/spinn_front_end_common/utilities/utility_calls.py#L103

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
```diff
@@ -29,96 +28,149 @@
...
+    :param ~.Placement placement: The placement to regenerate
+    :param str data_dir: A place to use to write data to
+    :return: Whether the data was regenerated or not
+    :rtype: bool
+    """
+    vertex = placement.vertex
+
+    # If the vertex doesn't regenerate, skip
+    if not isinstance(vertex, AbstractRewritesDataSpecification):
+        return False
```
Christian-B, line 75: Why return a Value if none ever reads it

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
```diff
@@ -29,96 +28,149 @@
...
+        # Ignore the deletion of files as non-critical
+        pass
+
+    # Read the region table for the placement
+    regions_base_address = txrx.get_cpu_information_from_core(
+        placement.x, placement.y, placement.p).user[0]
+    start_region = get_region_base_address_offset(regions_base_address, 0)
+    table_size = get_region_base_address_offset(
+        regions_base_address, MAX_MEM_REGIONS) - start_region
+    ptr_table = numpy.frombuffer(txrx.read_memory(
```
Christian-B, line 107: You set a value and the overwrite that value in the next line!

Christian-B (changes requested):
see comments

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
rowleya: Ah I should have been clear that I wanted to use this in a different branch; this means that the function can be used in an extension (which I am doing elsewhere).

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
rowleya: True, but the function takes the ptr_table generated in this line in...

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
rowleya: These lines turned out to be unused in master, precisely because they are used in get_report_writer instead!

Christian-B: ok

spinn_front_end_common/interface/interface_functions/dsg_region_reloader.py
Christian-B: Stand alone it would then be cleaner if you move the create ptr_table into the second function.

If however you intend to use the second function elsewhere than fine.

rowleya: It is only this function that is visible to the outside, so I have moved the code for getting the pointer table into the function for fixing it (and renamed it of course)

Christian-B:
ok to merge if ready

PR #989, opened 2022-10-11
more flexibility for when tests interact

Christian-B:
related to https://github.com/SpiNNakerManchester/SpiNNUtils/pull/195 but can be merged separately.

If all test run together on some systems the tests can interact.
For example on Windows the log appears to go into a different thread and does not happen until the next test is already running.

tested by:


Christian-B:
Brought into another pr

PR #990, opened 2022-10-12
Split provenance

Christian-B:
This PR Splits provenance database into two.

Some tables including:
version information
Timer information
log messages

Have been moved to global provenance database directly under the timestamp directory (parallel to run_1)

ProvenanceReader and ProvenanceWriter manage which data goes where,
Both can be setup with none standard database files with only the one used for calls needed
The run_query method has been replaced with two calls as it needs to know which database to ask.
(Recommended is to add Semantic sugar methods remains!)

Added a ProvenanceBase so there is only one place where the paths to the 2 sql files are created.

-
_TestExecutableFinder now extends ExecutableFinder  to avoid issues with Test interactions.

some provenance database test also changed for case of test interactions

Most be done at the same time as:

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/153


rowleya:
Merge when testing done

PR #995, opened 2022-10-21
Timer report

Christian-B:
This PR has.

1. Fixes to the timings in ASB
   - add FecTimer.start_category(TimerCategory.GET_MACHINE, True) to actully record the machine on time
   - added the DATA_GENERATION category 
   
2. Adds a timer reports
   - run at the end of stop so it report over multiple resets (after a merge with https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/990)
   - Report saved in timestamp directory as it covers all runs and all resets (or will after above merge)
   - Can also be run standalone over a previous run database (One created after the fixes above have gone in)

Jenkins tested with report always on but suggest turning it off by default. Especially as it can be easily done later.

Added tests to demonstrate stand alone

There is a microcircuit_model PR which runs and reports the timer reports in jenkins.
https://github.com/SpiNNakerManchester/microcircuit_model/pull/21


rowleya:
I could be wrong but I think we have this already...

Christian-B:
What we have is a lot less than this

PR #997, opened 2022-10-25
new buffer database with each Reset

Christian-B:
This pr creates a new buffer sqlite3 file with each reset rather than delete it.

Also includes preparation work for merging the "buffer" and "provenance" sqlite3 files

changes included.
View tracking the reset number

BufferManager instead holding a database it uses the with pattern
On reset the old sqlite3 is no longer deleted

BaseDatabase class created which will hold code shared with the ProvenanceReader/ProvenanceWriter when they later share the same sqlite3 file

https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/628

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/159

Replaces:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/987


Christian-B:
Now part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1015

PR #999, opened 2022-10-27
Merge sqlite databases

Christian-B:
Builds on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/997 and https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/998

buffer.sqlite3 and provenance.sqlite3 are now combined into data.sqlite3

PR #1001, opened 2022-10-27
Time report2

Christian-B:
Combines https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/995
and https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/999

there is now a single timer report which works even if there are hard resets.

Related PRs in other repositories. (may be part of the ones being combined)

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1240
https://github.com/SpiNNakerManchester/microcircuit_model/pull/21

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/157


Christian-B:
part already merged in so the combine does not make sense

PR #1004, opened 2022-11-04
Reduce overhead data spec

rowleya:
Minor fix: rewrite data spec should be done if required!

Note part of reduce_overhead_data_spec as this requires this fix to work correctly, but can be merged independently.


andrewgait (approved):
This also looks fine.

PR #1025, opened 2023-02-28
Fix multi output

rowleya:
The I/O database key mapper is filled from many places.  Sometimes a vertex is in more than once place.  This makes a set rather than a list to avoid issues.


andrewgait (approved):
I was about to say this might not be such a good idea for machine vertices in general, but as this is only for LPG machine vertices, it's fine.

PR #1026, opened 2023-02-28
Additional router provenance

rowleya:
Add additional router provenance field to identify the source of blockages


andrewgait (approved):
Always good to have an extra feature like this.

PR #1027, opened 2023-03-03
Fix spif compress 2

rowleya:
Improvements to compression speed through the use of DMAs.

A quick check says this also sneaks in a minor fix to the database so that it only gets keys for outgoing vertices when using live output (which makes sense).


andrewgait (approved):
Looks good to me.

PR #1028, opened 2023-03-07
Report the binary file in use

rowleya:
Adds the reporting of which binary was used on which vertex, which can be very useful in multi-binary models!


dkfellows:
Do we have any non-virtual machine vertices that do not implement `AbstractHasAssociatedBinary`? Isn't implementing that an absolute requirement for being deployed to the machine at all?

rowleya:
> Do we have any non-virtual machine vertices that do not implement `AbstractHasAssociatedBinary`? Isn't implementing that an absolute requirement for being deployed to the machine at all?

Yes, pretty much - I can see that the else case here is sort-of not needed.  Ideally we would add the get_binary methods to the machine vertex, and return None by default I suppose, thus negating the need for AbstractHasAssociatedBinary at all, or alternatively have a OnMachineVertex or something.  That seems like the job for another PR though...

PR #1032, opened 2023-03-17
turn of validation of files used by JavaCaller

Christian-B:
Due to https://github.com/python-jsonschema/jsonschema/issues/1059 jsonschema validation sometime fails.

This pr removes validation from normal run.
This will also provide a very minor speedup by removing an unlikely to be failed safety test

But keeps it when writing the json files as reports.
For example in debug mode which is tested by
spynnaker_integration_tests/test_debug_mode

### DIFF
--- a/spinn_front_end_common/interface/java_caller.py
+++ b/spinn_front_end_common/interface/java_caller.py
@@ -170,7 +170,8 @@ class JavaCaller(object):
         :return: the name of the file containing the JSON
         """
         if self._machine_json_path is None:
-            self._machine_json_path = write_json_machine(progress_bar=False)
+            self._machine_json_path = write_json_machine(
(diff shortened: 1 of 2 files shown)


dkfellows:
This looks OK in as far as it goes, but I want to know why we are having schemas cross threads. Is this just in testing?

Christian-B:
So far the error has only been spotted on Jenkins which is multi threaded.

But even there it happens in one or tow of several hundred runs so hard to be sure it could not happen elsewhere.

dkfellows:
Interesting. We're not passing any schema objects between threads ourselves. That's being done by some sort of cache within jsonschema.

dkfellows (approved):
Acceptable workaround; jsonschema is more broken than I thought right now.

PR #1034, opened 2023-03-23
Receiver ping

rowleya:
Makes it possible for an external receiver sitting behind a firewall to continuously receive packets.  This is done by repeatedly setting the tag (once every 10 seconds) which keeps the firewall port open.


andrewgait (approved):
Looks fine to me - I was just wondering whether you'd used 10 seconds based on some testing, or if that was the first thing you tried?

PR #1035, opened 2023-03-23
Doc polish

dkfellows:
This is a round of doc polishing and fixing of error messages and so on.

See SpiNNakerManchester/SpiNNUtils#223


dkfellows:
Still an issue with not picking up links into PACMAN.

dkfellows:
That's much better now!

PR #1041, opened 2023-04-13
Enable spell checking of docstrings

dkfellows:
See:
* https://github.com/SpiNNakerManchester/SupportScripts/pull/54
* https://github.com/SpiNNakerManchester/SpiNNUtils/pull/225


dkfellows:
That looks good enough to me.

PR #1043, opened 2023-04-17
Gfe live fixes

rowleya:
Makes it possible to to GFE live input and output using Machine Vertices rather than Application Vertices (Machine Vertices are wrapped in App vertices but not in a way that worked with the live IO).


spinn_front_end_common/utilities/database/database_writer.py
```diff
@@ -267,6 +270,21 @@ def create_atom_to_event_id_mapping(self, machine_vertices):
...
 
+    def _is_machine_lpg(self, vertex):
+        return (isinstance(vertex, LivePacketGatherMachineVertex) and
+                not isinstance(vertex.app_vertex, LivePacketGather))
+
+    def _is_machine(self, edge):
+        return (isinstance(edge.pre_vertex, AbstractOneAppOneMachineVertex) and
+                isinstance(edge.post_vertex, AbstractOneAppOneMachineVertex))
+
+    def _get_machine_lpg_mappings(self, part):
```
Christian-B, line 281: I think this will read a lot cleaner if the three functions are combined.

It also avoids directly doing edge.post_vertex.machine_vertex..app_vertex

Christian-B (commented):
look ok just a but heard to read due to the many tiny methods

spinn_front_end_common/utilities/database/database_writer.py
rowleya: Done

PR #1044, opened 2023-04-17
Better error handling

rowleya:
Make one place for error handling and use it.

Also adds a couple of progress bars to application run as some things can take a while in there...


rowleya:
Currently not quite working it seems, so don't merge yet...

PR #1048, opened 2023-04-20
ExecutableType moved to SpiNNMan

Christian-B:
Create a stub to allow the moved ExecutableType  to still be imported from FrontendCommon

See: https://github.com/SpiNNakerManchester/SpiNNMan/pull/323

### DIFF
--- a/fec_integration_tests/interface/interface_functions/test_front_end_common_load_executable_images.py
+++ b/fec_integration_tests/interface/interface_functions/test_front_end_common_load_executable_images.py
@@ -17,10 +17,9 @@ from collections import defaultdict
 from spinn_utilities.overrides import overrides
 from spinnman.transceiver import Transceiver
 from spinnman.model import ExecutableTargets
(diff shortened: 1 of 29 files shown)


Christian-B:
Using only the stud from spinn_front_end_common.utilities.utility_objs import ExecutableType
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/move_ExecutableType/1/pipeline

rowleya:
Warning: the warning might occur even when you don't directly import the class.

If this isn't the case, it might be easier to just do "from spinnman.model.enums import ExecutableType", which will immediately make it available here.

Christian-B:
> Warning: the warning might occur even when you don't directly import the class.
> 
> If this isn't the case, it might be easier to just do "from spinnman.model.enums import ExecutableType", which will immediately make it available here.

Agreed removed.

PR #1055, opened 2023-04-28
Checksum itcm

rowleya:
Adds checksumming of ITCM during run.  Note currently causes ITCM overrun on some binaries!


rowleya:
This now compiles correctly on examples

PR #1058, opened 2023-05-03
removes requires_data_generation

Christian-B:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1338 removes the only use of requires_mapping as it is currently not working.

With that use case gone the whole requires mapping can go.

This simplifies ASB which no longer has to consider case bewteen a hard and soft reset. (no mapping but some load)

This then allows:
https://github.com/SpiNNakerManchester/SpiNNUtils/pull/230


Christian-B:
Did not work as expected.

PR #1060, opened 2023-05-10
fixes the edge_routing_info.rpt for virtual vertices

Christian-B:
fixes the edge_routing_info.rpt for virtual vertices

tested by:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1340

and:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/201


spinn_front_end_common/utilities/report_functions/reports.py
```diff
@@ -209,10 +210,11 @@ def _write_one_router_partition_report(f, partition):
     routing_infos = FecDataView.get_routing_infos()
     for edge in partition.edges:
         for m_vertex in outgoing:
-            source_placement = FecDataView.get_placement_of_vertex(m_vertex)
+            if isinstance(m_vertex, AbstractVirtual):
+                continue
```
rowleya, line 214: I think this means you won't get any routes starting at a virtual vertex, which seems like an ommision...  Note that virtual vertices have keys and masks too I believe...

Christian-B: Agree better to report something than nothing.

But this exposes that in [sPyNNaker/unittests/test_sata_connectors/test_sata_2_different_boards_valid_board_address.py](https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/unittests/test_sata_connectors/test_sata_2_different_boards_valid_board_address.py)

The second source ends up with no entry

Which does not happen in 
https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/unittests/test_spinnaker_link_connectors/test_spinnaker_link_2_different_boards_valid_board_address.py

spinn_front_end_common/utilities/report_functions/reports.py
```diff
@@ -677,31 +675,29 @@ def generate_comparison_router_report(compressed_routing_tables):
...
     :rtype: tuple(str, int)
     """
     # Create text for starting point
     machine = FecDataView.get_machine()
-    source_vertex = source_placement.vertex
     text = ""
     if isinstance(source_vertex, MachineSpiNNakerLinkVertex):
-        text = "        Virtual SpiNNaker Link on {}:{}:{} -> ".format(
-            source_placement.x, source_placement.y, source_placement.p)
+        text = "        Virtual SpiNNaker Link on -> "
```
rowleya, line 688: I suspect the fix here is to use x, y, slink in the formatted output to match the placement, as otherwise it isn't too clear where the link comes from

spinn_front_end_common/utilities/report_functions/reports.py
```diff
@@ -677,31 +675,29 @@ def generate_comparison_router_report(compressed_routing_tables):
...
         slink = machine.get_spinnaker_link_with_id(
             source_vertex.spinnaker_link_id)
         x = slink.connected_chip_x
         y = slink.connected_chip_y
     elif isinstance(source_vertex, MachineFPGAVertex):
-        text = "        Virtual FPGA Link on {}:{}:{} -> ".format(
-            source_placement.x, source_placement.y, source_placement.p)
+        text = "        Virtual FPGA Link on -> "
```
rowleya, line 694: Similar to previous comment - use x, y, flink to fill in the details here

rowleya (dismissed):
Minor change requested

spinn_front_end_common/utilities/report_functions/reports.py
Christian-B: Not sure I agree as the slink x, and y are part of the report but added as requested.

PR #1064, opened 2023-05-18
Make everything work over dimensions

rowleya:
Part of SpiNNakerManchester/sPyNNaker#1345.

Removes the bitfield compression as no longer works well with multiple dimensions.  Also updates other things for multi-dimensional use.


Christian-B (approved):
looks ok

PR #1065, opened 2023-05-25
Remove the potential collision report

dkfellows:
It was broken. It could not produce correct results. It was disabled. It was unreachable.

It is gone.


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -1360,7 +1347,6 @@ def _do_mapping(self, total_run_time):
         self._report_routers()
         self._report_router_summary()
         self._json_routing_tables()
-        # self._report_router_collision_potential()
```
dkfellows, line 1363: This is why things were unreachable

spinn_front_end_common/utilities/report_functions/router_collision_potential_report.py
```diff
@@ -1,81 +0,0 @@
...
-
-
-def _get_collisions_with_other_router(
-        x, y, collision_route, router_tables_by_partition, n_key_map):
-    total = 0
-    for partition in router_tables_by_partition.get_entries_for_router(x, y):
-        entry = router_tables_by_partition.get_entry_on_coords_for_edge(
-            partition, x, y)
-        if collision_route in entry.link_ids:
-            total += n_key_map.n_keys_for_partition(partition)
```
dkfellows, line 80: This is why it was broken... because `n_key_map` had to be a plain `dict` at this point. 🤦

PR #1069, opened 2023-06-05
Major cleanup/ reduction of the DataSpecification

Christian-B:
Moves all DS code but DataType to FrontEndCommon
DataType to be moved in a seperate PR due to the amount of places it is used.
All other DS code deleted from DataSpecification see: https://github.com/SpiNNakerManchester/DataSpecification/pull/151

Must be done at the same time as https://github.com/SpiNNakerManchester/DataSpecification/pull/151
 https://github.com/SpiNNakerManchester/sPyNNaker/pull/1352 and https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/884
there is a microcircuit_model branch but that just enables more report so can be removed
Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/205
(This appears to fail more often than master (maybe as it does not use java??)

AbstractRewritesDataSpecification changed to having to be an AbstractGeneratesDataSpecification
AbstractGeneratesDataSpecification changed to having to be an AbstractHasAssociatedBinary
This bacause the code applied only makes sense if this holds (it does by implementation in master)

ReferenceContext replaced with a ...View get_next_ds_references

constants from DS constants that are still needed have been moved to FEC constants
ds cfg merged into fec cfg
Instead of the many DS exceptions there is now just DataSpecException
 (as well as DsDatabaseException being used more often)

The DS data from ds.write calls is now held by the DataSpecificationGenerator and then written to the database ASAP so the
DataRowWriter is no longer needed.

Most of the functionality from dsg_region_reloader has been moved to DataSpecificationReloader

The DataSpecificationReloader and DataSpecificationGenerator share code which is now in 
DataSpecificationBase

The API of DataSpecificationGenerator has been significantly reduced as much of it was never used.
(or removed in https://github.com/SpiNNakerManchester/sPyNNaker/pull/1344)

The remaining functionality of DSG reserve regions,, write data, set references is now handled by writing this information directly to the database in several tables and rows rather than previous single bitarray with instructions and data mingled together

The  DataSpecificationReloader no longer reads the pointer table off the core instead uses the ds.sqlite3 to obtain the data to check that reserve regions sizes are unchanged. (Note reload reserve region calls are no longer needed but do provide a safety check)

DsSqlliteDatabase and dse.sql have been significantly changed and could almost be consider new.
Some methods renamed for clarity
A lot of the safety code previously in DataSpecificationExecutor is now handled by the database
Computing region pointer /start address now done by the database
reference matching is now done by the database

DataSpecificationExecutor is completely removed as no longer needed

host_execute_data_specification.py significantly simplified
The algorithm is now
- for each code 
     -  malloc the region 
     - save the base address to the which calculates the start of each region
- for each core
     - write each region
          - build the pointer table while doing so
     - fill in reference addresses
     - write the pointer table
     
(this causes a minor change that the pointer table is written to core after (time not space) the regions)

Added new unittest for database (call through DataSpecificationGenerator  whenever possible)
Update older tests unless they really did not make sense any more (than deleted)

### DIFF
--- a/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
+++ b/spinn_front_end_common/interface/interface_functions/spalloc_allocator.py
@@ -146,7 +146,7 @@ class SpallocJobController(MachineAllocationController):
 
     @overrides(AbstractMachineAllocationController.open_eieio_listener)
     def open_eieio_listener(self):
-        return self._job.open_listener_connection()
+        return self._job.open_eieio_listener_connection()
 
     @property
     @overrides(AbstractMachineAllocationController.proxying)
--- a/spinn_front_end_common/utilities/connections/live_event_connection.py
+++ b/spinn_front_end_common/utilities/connections/live_event_connection.py
@@ -287,7 +287,7 @@ class LiveEventConnection(DatabaseConnection):
         if self.__sender_connection is None:
             job = db.get_job()
             if job:
-                self.__sender_connection = job.open_listener_connection()
+                self.__sender_connection = job.open_eieio_listener_connection()
             else:
                 self.__sender_connection = EIEIOConnection()
         for label in self.__send_labels:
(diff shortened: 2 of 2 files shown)


dkfellows (approved):
If it works in integration test, I'm happy.

PR #1086, opened 2023-07-10
Defaults from cfg

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/215

Config_holder has a stub that of now just sets spin1

To be determined how to distinguish in a cfg which board type to be expected.


Christian-B:
https://github.com/SpiNNakerManchester/SpiNNMachine/pull/217 was a better method

PR #1104, opened 2023-08-18
hide get_cpu_infos

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNMan/pull/358

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -2268,8 +2268,8 @@ class AbstractSpinnakerBase(ConfigHandler):
         if not unsuccessful_cores:
             for executable_type, core_subsets in \
                     self._data_writer.get_executable_types().items():
-                failed_cores = transceiver.get_cpu_infos(
-                    core_subsets, executable_type.end_state, False)
+                failed_cores = transceiver.get_cores_in_states(
+                    core_subsets, executable_type.end_state)
                 for (x, y, p) in failed_cores:
                     unsuccessful_cores.add_processor(
                         x, y, p, failed_cores.get_cpu_info(x, y, p))
@@ -2301,8 +2301,8 @@ class AbstractSpinnakerBase(ConfigHandler):
             # Extract any written provenance data
             try:
                 transceiver = self._data_writer.get_transceiver()
-                finished_cores = transceiver.get_cpu_infos(
-                    non_rte_core_subsets, CPUState.FINISHED, True)
+                finished_cores = transceiver.get_cores_in_states(
+                    non_rte_core_subsets, CPUState.FINISHED)
                 finished_placements = Placements()
                 for (x, y, p) in finished_cores:
                     try:
--- a/spinn_front_end_common/interface/interface_functions/application_finisher.py
+++ b/spinn_front_end_common/interface/interface_functions/application_finisher.py
@@ -62,16 +62,16 @@ def application_finisher():
                 f"{total_processors} processors went into an error state "
                 "when shutting down")
 
-        successful_cores_finished = txrx.get_cpu_infos(
-            all_core_subsets, CPUState.FINISHED, True)
+        successful_cores_finished = set(txrx.get_cores_in_states(
+            all_core_subsets, CPUState.FINISHED))
 
         txrx.send_signal(app_id, Signal.SYNC0)
         txrx.send_signal(app_id, Signal.SYNC1)
 
         for core_subset in all_core_subsets:
             for processor in core_subset.processor_ids:
-                if not successful_cores_finished.is_core(
-                        core_subset.x, core_subset.y, processor):
+                if (core_subset.x, core_subset.y, processor) not in \
+                        successful_cores_finished:
                     txrx.update_provenance_and_exit(
                         core_subset.x, core_subset.y, processor)
         time.sleep(0.5)
--- a/spinn_front_end_common/interface/interface_functions/chip_provenance_updater.py
+++ b/spinn_front_end_common/interface/interface_functions/chip_provenance_updater.py
@@ -61,15 +61,17 @@ class _ChipProvenanceUpdater(object):
             left_to_do_cores,
             "Forcing error cores to generate provenance data")
 
-        cpu_infos = self.__txrx.get_cpu_infos(
+        bad_cores = self.__txrx.get_cores_in_states(
             self.__all_cores,
-            [CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG, CPUState.IDLE],
-            True)
-        error_cores = cpu_infos.infos_for_state(CPUState.RUN_TIME_EXCEPTION)
-        watchdog_cores = cpu_infos.infos_for_state(CPUState.WATCHDOG)
-        idle_cores = cpu_infos.infos_for_state(CPUState.IDLE)
+            [CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG, CPUState.IDLE])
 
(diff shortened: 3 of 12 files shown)


spinn_front_end_common/interface/abstract_spinnaker_base.py
```diff
@@ -2270,8 +2270,8 @@ def __recover_from_error(self, exception):
         if not unsuccessful_cores:
             for executable_type, core_subsets in \
                     self._data_writer.get_executable_types().items():
-                failed_cores = transceiver.get_cpu_infos(
-                    core_subsets, executable_type.end_state, False)
+                failed_cores = transceiver.get_cores_in_states(
```
rowleya, line 2273: Should this be "not in state" since it was include=False?

spinn_front_end_common/interface/interface_functions/application_finisher.py
```diff
@@ -62,16 +62,16 @@ def application_finisher():
...
 
         txrx.send_signal(app_id, Signal.SYNC0)
         txrx.send_signal(app_id, Signal.SYNC1)
 
         for core_subset in all_core_subsets:
             for processor in core_subset.processor_ids:
-                if not successful_cores_finished.is_core(
-                        core_subset.x, core_subset.y, processor):
+                if (core_subset.x, core_subset.y, processor) not in \
+                        successful_cores_finished:
```
rowleya, line 74: Is that change necessary?  It seems that the function get_cores_in_states returns the same information as get_cpu_infos, but it has now been put in a set instead of using the built-in set...

spinn_front_end_common/interface/interface_functions/chip_provenance_updater.py
```diff
@@ -59,15 +59,17 @@ def _run(self):
...
-        if error_cores or watchdog_cores or idle_cores:
+            [CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG, CPUState.IDLE])
+
+        if bad_cores:
+            error_cores = self.__txrx.get_cores_in_states(
+                self.__all_cores, CPUState.RUN_TIME_EXCEPTION)
+            watchdog_cores = self.__txrx.get_cores_in_states(
+                self.__all_cores, CPUState.WATCHDOG)
+            idle_cores = self.__txrx.get_cores_in_states(
+                self.__all_cores, CPUState.IDLE)
```
rowleya, line 72: Similarly to a previous comment, get_cores_in_states returns the same information as get_cpu_infos so can't you still do the same thing rather than get them all again?

spinn_front_end_common/utilities/emergency_recovery.py
```diff
@@ -34,36 +34,23 @@ def _emergency_state_check():
...
         if rte_count or watchdog_count:
-            states = txrx.get_cpu_infos(
-                None, [CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG], True)
+            bad_cores = txrx.get_cores_in_states(
+                None, [CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG])
             logger.warning(
                 "unexpected core states (rte={}, wdog={})",
                 rte_count, watchdog_count)
-            logger.warning(states.get_status_string())
+            logger.warning("cores {}", set(bad_cores))
```
rowleya, line 42: I think this can still get the status string?

spinn_front_end_common/utilities/emergency_recovery.py
```diff
@@ -34,36 +34,23 @@ def _emergency_state_check():
...
-                        chip.x, chip.y, p)
-                    if info.state in (
-                            CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG):
-                        infos.add_processor(chip.x, chip.y, p, info)
-                except Exception:
-                    errors.append((chip.x, chip.y, p))
-        if len(infos):
-            logger.warning(infos.get_status_string())
+        errors = txrx.get_cores_in_states(
+            CPUState.RUN_TIME_EXCEPTION, CPUState.WATCHDOG)
```
rowleya, line 47: This is a little more subtle - I think the idea was in the case of an error to go chip-by-chip to avoid getting a load of errors all at once

rowleya: Also note errors is when things go wrong in reading the state

spinn_front_end_common/utilities/emergency_recovery.py
```diff
@@ -34,36 +34,23 @@ def _emergency_state_check():
...
         if len(errors) > 10:
             logger.warning(
                 "Could not read information from {} cores", len(errors))
         else:
             logger.warning(
-                "Could not read information from cores {}", errors)
+                "Could not read information from cores {}", set(errors))
```
rowleya, line 53: See above about the change in meaning of errors

spinn_front_end_common/utilities/system_control_logic.py
```diff
@@ -163,13 +163,12 @@ def _load_application(executable_targets, app_id):
     # Check that the binaries have reached a wait state
     count = transceiver.get_core_state_count(app_id, CPUState.READY)
     if count < executable_targets.total_processors:
-        cores_ready = transceiver.get_cpu_infos(
-            executable_targets.all_core_subsets, [CPUState.READY], False)
+        cores_ready = transceiver.get_cores_in_states(
+            executable_targets.all_core_subsets, [CPUState.READY])
```
rowleya, line 167: Interesting - in master this seems to hold "cores_ready" but actually it is "cores_not_ready"...  I think it is meant to show the status of things that are not ready though... (and then it should be able to use the status string)

spinn_front_end_common/utility_models/data_speed_up_packet_gatherer_machine_vertex.py
```diff
@@ -1000,10 +1000,11 @@ def unset_cores_for_data_streaming(self):
                 FecDataView.iterate_monitors())
             try:
                 transceiver = FecDataView.get_transceiver()
-                error_cores = transceiver.get_cpu_infos(
-                    core_subsets, {CPUState.RUNNING}, False)
+                error_cores = transceiver.get_cores_in_states(
+                    core_subsets, {CPUState.RUNNING})
```
rowleya, line 1004: Should be not in state

rowleya (changes requested):
Quite a lot of these changes are hard to test because they only happen on errors which then haven't happened in testing!  I believe that get_cores_in_states and get_cores_not_in_states still return the same thing as get_cpu_infos, so it shouldn't be necessary to wrap them in a set...

spinn_front_end_common/interface/abstract_spinnaker_base.py
Christian-B: agreed

spinn_front_end_common/interface/interface_functions/application_finisher.py
Christian-B: The difference is that get_cpu_infos promises to return a spinnman.model.CPUInfos while get_cores_(not_)in_states only contracts to return an iterable(int, int, int) xyp.   This gives Spinn2 the freedom the have a different "CPUInfos" model

spinn_front_end_common/utilities/emergency_recovery.py
Christian-B: Yes the idea was to go chip by chip. BUT txrx.get_cpu_information_from_core did not support that.
Instead it gets all states and then returns the one wanted.

Christian-B:
Replaced with https://github.com/SpiNNakerManchester/SpiNNMan/pull/366

PR #1109, opened 2023-09-14
spalloc_machine may be none

Christian-B:


Christian-B:
spotted and tested locally

PR #1111, opened 2023-09-19
use one transaction for multiple calls

Christian-B:
This is a cherry pick and simplification of idea of using a db transaction longer from https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1062

As ProvenanceWriter is designed to be used as a ContextManager (using with)
context_entered starts a transaction
exit commits or rolls back

This depended on the SQLiteDB.connection property to expose the database in a way we did not really intend.
Instead
SQLiteDB 
- no longer has a connection property
- now has start_transaction, commit, rollback and cursor methods
- check to make sure cursor only happens with an open transaction and close only without an open transaction
- has a more light weight inner class _CursorWrapper which does not enter and exit the database

Unit test where added to show that
- commit and rollback are safe even when not expected.
- cursor and the ProvenanceWriter methods that depend on it will fail outside of a with. (id no open transaction)

### DIFF
--- a/spinn_front_end_common/interface/provenance/provenance_writer.py
+++ b/spinn_front_end_common/interface/provenance/provenance_writer.py
@@ -56,7 +56,7 @@ class ProvenanceWriter(BaseDatabase):
         :param str description: Type of value
         :param float the_value: data
         """
-        with self.transaction() as cur:
+        with self.cursor() as cur:
             cur.execute(
                 """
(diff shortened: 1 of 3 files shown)


dkfellows (approved):

Christian-B:
Done as https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1112 instead

PR #1113, opened 2023-09-25
Convert our use of SQLite to use RETURNING clauses

dkfellows:
This changes our Python code to use a `RETURNING` clause instead of using the `lastrowid` property, mirroring a change already in the Java code.

Note that this requires that we are running against a new enough Python to support `RETURNING` clauses. I don't know if 3.8 is new enough.


Christian-B:
This looks good assuming it works.

Can we also do something similar with rowcount

May also be an idea to convert the DatabaseReader use of fetchone

PR #1148, opened 2023-12-08
Simplify data in

rowleya:
Removes the complicated data in protocol and replaces it with a much simpler one!  This uses the same "multicast paths" on the machine, but instead of keeping track of packets and then retransmitting them in blocks, it instead uses normal SCP messages to send the data, which is then expanded to multicast on the machine as before.  This appears to run at the same speed as the other protocol, but without the complicated code...


Christian-B:
I ran a the self synfire with 100,000 neurons with a giant fromlist connector master and this branch.  Load Application data specification master 3,639,730 vs here. 7,001,228.  This is in Python!

PR #1159, opened 2024-03-05
use the version ncores per chip map

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/238


Christian-B:
This is the wrong fix.

PR #1163, opened 2024-03-07
Live tweaks

rowleya:
Fixes to make live output work with Machine vertices


rowleya:
It seems like the wrong fix was being done here.  This has been corrected.  The code that already existed already looked for LPG Machine vertices that had been added through the GFE, so the extra here previously was unnecessary.  The FEC code has been updated to account for the fact that no machine vertices should be added through the FECDataView in any case.  The other changes are still relevant and help with a few issues though.

PR #1199, opened 2024-07-10
switch between transceiver and monitor based on data size

Christian-B:


Christian-B:
Experiment with cutoff size show that not using the monitor is significantly better with single packet regions.

The monitor is considerably better with Very large (1,000,000 bytes +) where the non monitor took at least twice as long.

PR #1207, opened 2024-08-27
Simplfy the data.sqlite3

Christian-B:
https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/1172
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1479
tested by 
https://github.com/SpiNNakerManchester/IntegrationTests/pull/288

---
In master on a second run without reset data is split between region and region_extra
So two tables holding near the same data.

In this PR.
table region_data table replaces both tables

supported by
extraction and setup tables.

between the three tables for each core extraction the database holds
x, y, processor, local_region_index, content, content_len, run_timestep, run_time_ms, n_run, n_loop, extraction_time

---

This now writes a region data row for every recording region every extraction (run/auto pause loop) even if the data read is empty..  This allows distinguish between the correct data read but empty and the error of never reading the region.

Also on clear the record is not removed just changed to empty and the missing_data set to 2

---

Getting data raises clearer exceptions.
- If a virtual Machine is being used missing record records are allowed
- if missing and incorrect type NotImplementedError
- if missing and correct type BufferedRegionNotPresent saying either region not read or region not set to record

---
The BufferManager methods that Extract data from cores have been renamed with "extract" rather than get or retreive

### DIFF
--- a/spinn_front_end_common/interface/abstract_spinnaker_base.py
+++ b/spinn_front_end_common/interface/abstract_spinnaker_base.py
@@ -2041,7 +2041,7 @@ class AbstractSpinnakerBase(ConfigHandler):
             if timer.skip_if_virtual_board():
                 return
             bm = self._data_writer.get_buffer_manager()
-            bm.get_placement_data()
+            bm.gather_placement_data()
 
     def _do_extract_from_machine(self) -> None:
         """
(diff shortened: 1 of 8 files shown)


rowleya (approved):
This looks fine, but the multiple entries may have had something to do with a size limit on the blob...

PR #1209, opened 2024-09-10
recording data fixes

Christian-B:
Followup to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1207 and https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1202

- Add a method to FecData say if this is the last/or only auto pause loop of a run
- Only extract from AbstractReceiveRegionsToHost on the last/only loop
- Change BufferManager.get_placement_data so that for AbstractReceiveRegionsToHost regions it returns ONLY the last data extracted
- Removed the BufferManager.get_last_placement_data method as it is no longer needed.

To do this the database now tracks if the regions is_recording
- True for regions where the data is appended on each loop
- False for regions where the data is just the state at the end of the run
- Cleaned up some columns that should have been removed in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1207 

The Python tells the Java which placements to extract on each loop!

https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/1189
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1483

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/290

### DIFF
--- a/spinn_front_end_common/data/fec_data_view.py
+++ b/spinn_front_end_common/data/fec_data_view.py
@@ -94,6 +94,7 @@ class _FecDataModel(object):
         "_live_packet_recorder_params",
         "_live_output_vertices",
         "_live_output_devices",
+        "_n_run_steps",
         "_next_sync_signal",
         "_next_ds_reference",
         "_none_labelled_edge_count",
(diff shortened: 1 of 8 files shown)


rowleya:
A thought on this is that if we are now noting which regions are "recorded data" and which are not, we could use that to avoid region id clashing i.e. make it that the unique thing in the database is x, y, p, region_id, is_recorded.  Then get_data_by_placement could also pass in whether it is reading a recorded data region or not.

PR #1223, opened 2024-10-22
Allow run forever pause

rowleya:
Allow run_forever to be called multiple times in a row.  This still doesn't allow run_forever followed by a fixed time run or vice versa as the timing is then uncertain.  Run forever multiple times can make sense to some simulations though.


Christian-B (approved):
looks fine.

An example would not hurt.

Jenkins run is never a bad idea too

PR #1224, opened 2024-10-25
Run forever fixes

Christian-B:
Changes how run forever works with recording.

automated auto pause resume loops have been removed.
- You could not break out of these with a stop
- No know use case actually ran a second loop

Now run forever with record is a run and trust stop will be called on time
- Record space is set as large as possible
- The is a log to say how long the run can be trusted
- A single run IE no loops is tarted

These changes allow the ChipPowerMonitors to be added

In this PR the compute energy usage INCORRECTLY assumes the runtime is 0

fixes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1228 https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1229

### DIFF
--- a/c_common/front_end_common_lib/include/simulation.h
+++ b/c_common/front_end_common_lib/include/simulation.h
@@ -66,7 +66,9 @@ typedef enum simulation_commands {
     IOBUF_CLEAR = 9,
     //! Asks the application to pause.  This relies on the application using
     //! simulation_is_finished which can then handle the pause status better.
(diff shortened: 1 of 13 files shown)


rowleya (approved):
This all looks fine, once it is passing

Christian-B:
What is still needed and not done elsewhere is in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1234

PR #1231, opened 2024-11-06
Get current time support

rowleya:
Add support to get the current time of simulation from cores, and then setting the current run time when running forever or until stop.

Also needs merging at same time:
 - SpiNNakerManchester/SpiNNMan#426
 - SpiNNakerManchester/sPyNNaker#1503

Tested by:
 - SpiNNakerManchester/IntegrationTests#302


rowleya:
Note test failures are due to usual noise, not related to this branch now

PR #1236, opened 2024-11-11
Better energy checkpointing

rowleya:
Another way to get energy report data without having to store checkpoints


Christian-B (approved):
approved when tested by code in another repository

PR #1242, opened 2024-11-13
Fix pause

rowleya:
Fix issues with pause.  Sometimes I see failure to read the next message because pause messages do return a response but it wasn't being read.  This makes pause request a full process, and then removes the need to do anything complex in the application_runner, except to wait for the end of simulation after sending the message.


Christian-B (approved):
approved assuming tested with spynnaker and TSP

PR #1246, opened 2024-11-19
Minor updates

rowleya:
Some fixes to avoid hazards in notification, especially when starting and stopping repeatedly

### DIFF
--- a/spinn_front_end_common/utilities/database/database_connection.py
+++ b/spinn_front_end_common/utilities/database/database_connection.py
@@ -144,7 +144,6 @@ class DatabaseConnection(UDPConnection):
             # Call the callback
             with DatabaseReader(database_path) as db_reader:
                 for db_callback in self.__database_callbacks:
-                    logger.info("Calling database callback {}", db_callback)
                     db_callback(db_reader)
         else:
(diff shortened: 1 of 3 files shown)


spinn_front_end_common/utilities/database/database_connection.py
```diff
@@ -144,6 +144,7 @@ def __read_db(self, toolchain_address: Tuple[str, int], data: bytes):
             # Call the callback
             with DatabaseReader(database_path) as db_reader:
                 for db_callback in self.__database_callbacks:
+                    logger.info("Calling database callback {}", db_callback)
```
Christian-B, line 147: Nested log not a good idea.

spinn_front_end_common/utilities/notification_protocol/notification_protocol.py
```diff
@@ -181,6 +179,9 @@ def __do_read_notify(self, database_path: Optional[str]) -> None:
         if self.__wait_for_read_confirmation:
             for c in self.__database_message_connections:
                 try:
+                    logger.info(
```
Christian-B, line 182: Nested log not a good idea.

Christian-B (changes requested):

PR #1249, opened 2024-12-06
Allow energy reading even if not fully accurate

rowleya:


Christian-B (changes requested):
correct idea.

Poor implementation.

Just check if using energy monitors and skip the whole try block!

Christian-B:
typing fixed by https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1253

Christian-B:
alternative merged

PR #1253, opened 2024-12-16
Typing for PACMAN

Christian-B:
part of https://github.com/SpiNNakerManchester/PACMAN/pull/585


Christian-B:
This is a BLOCKER so self approving

Issue #1260, opened 2025-01-28
Report don't use mode Production if accessing databases after end


Christian-B:
Mode production correctly deletes databases

It should be possible to detect mode production and that end has been called where a database file not found error is going to be called

Wrap that error in a better error is possible.

PR #1261, opened 2025-01-28
better file_not_found

Christian-B:
fixes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1260


Christian-B:
tested with  spynnaker_integration_tests/test_power_monitoring (mode production)

FileNotFoundError: In Mode production many files are deleted on end. This likely caused no such DB: .....

Issue #861, opened 2021-09-16
Use of Nuke Algorithms


Christian-B:
FindApplicationChipsUsed | NChipsUsed ... | No use found
PreAllocateForBitFieldRouterCompressor | PreAllocatedResources token GeneratedPreAllocatedResources.BitFieldRouterCompressor | No use found
RoutingCompressionChecker | None | No use found
SpYNNakerConnectionHolderGenerator | ConnectionHolders | No use found
SynapticMatrixReport | None | Broken
ValidRoutesChecker | None | No use found

Christian-B:
There is also BasicDijkstraRouting and NerRoute

Christian-B:
ZonedRoutingTableGenerator also not used

Christian-B:
WriteJsonMachineGraph also unused

Christian-B:
Also GraphEdgeWeightUpdater

Christian-B:
Check and still there and still unsused are:
FindApplicationChipsUsed
PreAllocateForBitFieldRouterCompressor
SpYNNakerConnectionHolderGenerator
SynapticMatrixReport
WriteJsonMachineGraph

Christian-B:
There are now PRs to fix or remove the Algorithm not already gone.

PR #1348, opened 2026-03-30
Validate routes uncompressed

Christian-B:
Changed to Debug so the Check can run as needed

related to https://github.com/SpiNNakerManchester/PACMAN/pull/627

part of the fix for https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/861


rowleya (approved):
Happy for this to be turned on once it does what it is supposed to do!

PR #1353, opened 2026-04-13
Energy Cleanup

Christian-B:
includes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1350

cleans up the energy report

The report was done after every run but contained a mix of reset level and global level data

- Move it to the reset and stop stage
  - So now includes energy time between run and reset or stop call
- Consistently use reset data or all data 

"loading_time" changed to "ds_time" to better show it is the Data Spec stage.

Instead of a mix of TimerCategory and TimerWork the Energy Only uses TimerCategory
Added a few new categorises as needed by the energy report
DATA_SPEC_OTHER was LOADING  (as it is the DataSpec stage)
DATA_SPEC_LOAD was  TimerWork.LOADING_DATA
DATA_SPEC_SYNAPSE was TimerWork.SYNAPSE
MACHINE_ON was GET_MACHINE.  (Also now always called)
MACHINE_OFF  new  in close_allocation_controller called during hard reset
EXTRACT_DATA was TimerWork.EXTRACT_DATA
(see https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1354)

new database method get_machine_on_by_reset()
Get the whole time the machine was on
- replaces summing all the various uninteresting categories
- excludes time until the machine was turned on
  -  either by get_machine or run/mapping stage

Chip Activity simplified to just the total.
- no need to create a large dict if all that was needed is the sum.
- Old method already summed all Cores on each Chip

monitor count fixed as it was wrong
Power monitor monitors removed only if used

n_reset moved to category_timer_provenance table and run_reset_mapping table (and a view removed)
-previous approach as wrong as reset number changed too soon

FecTimer now errors if same Category started twice

see https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1355
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1356
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1357


Christian-B:
tested by https://github.com/SpiNNakerManchester/IntegrationTests/actions/runs/24338414470

spinn_front_end_common/interface/interface_functions/compute_energy_used.py
```diff
@@ -48,48 +48,35 @@
...
-                TimerWork.EXTRACT_DATA)
-            run_loop_ms -= data_extraction_ms
-        expansion_ms = db.get_timer_sum_by_work(TimerWork.SYNAPSE)
-        loading_ms -= expansion_ms
+            data_extraction_ms = db.get_category_timer_sum(
+                TimerCategory.EXTRACT_DATA, n_reset)
+        else:
+            data_extraction_ms = 0
+        expansion_ms = db.get_category_timer_sum(
+            TimerCategory.DATA_SPEC_SYNAPSE, n_reset)
```
rowleya, line 79: Does this miss out the loading of other things e.g. routes and binaries?

Christian-B: No the rest of loading is:

data_loading_ms = db.get_category_timer_sum(
            TimerCategory.DATA_SPEC_LOAD, n_reset)

This replicates what master did

rowleya (approved):

PR #1399, opened 2026-09-02
wait once for just 45 seconds

Christian-B:
tests https://github.com/SpiNNakerManchester/SpiNNMan/pull/498

also reduces the test48 wait from 4 times 300 seconds to once 45 seconds


Christian-B:
self approving as just test config change

